PR #10, opened 2016-11-07
Add brightness and Cmus modules

paxy97:
Hi.
First, do you take pull requests at all and if yes what should I provide?
And secondly, should I add entries to the themes for this module?

### DIFF
--- a/bumblebee/modules/brightness.py
+++ b/bumblebee/modules/brightness.py
@@ -1,0 +1,22 @@
+import bumblebee.module
+
+def description():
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Hi,

absolutely, I appreciate pull requests a lot, thanks a lot for the contribution!

Regarding the themes: I want to make it as easy as possible for others to add their modules, so it's no problem at all if there's no theme entries, I'll just add them myself.

Hvala!

PR #13, opened 2016-11-17
Missing battery fix and gruvbox theme

paxy97:
I often remove my laptop battery when I'm at home and power it using the charger to increase my battery lifetime. The problem was that the battery indicator didn't like that since the file doesnt exist in /sys when no battery isnt installed, so I just added a check for that. I also added the gruvbox color scheme (which is my fav). Might add more if I find the time.

PS. Odakle si?

### DIFF
--- a/bumblebee/modules/battery.py
+++ b/bumblebee/modules/battery.py
@@ -1,5 +1,6 @@
 import datetime
 import bumblebee.module
+import os.path
 
 def description():
     return "Displays battery status, percentage and whether it's charging or discharging."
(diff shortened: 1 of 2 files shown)


paxy97:
Also, I'm sorry for not writing tests, I have never written a single one. But hopefully I'll look into it when I find the time, which is sadly scarce midsemester.

tobi-wan-kenobi:
Hi,

great, thanks a lot! The battery issue was something I was thinking about, but honestly, I forgot about it when committing :-) I like how you use a separate status to identify AC. When I get the chance, I'll update the other themes to incorporate that (and also update the gruvbox theme to use the cmus states I added). For me, too, this is a weekend-project, so I fully understand your mid-semester concerns ;-)

Also, no worries regarding the tests, I'm just getting started on those anyhow (as you can see from the fact that there's a bare handful...)

Austrijanac sam, ali ucim srpski/hravtski/bosanski (polako) ;-)

paxy97:
I added the new cmus support to the gruvbox theme, or am I missing something?

tobi-wan-kenobi:
Yes, you did. But I extended your cmus module to be able to toggle shuffle and repeat, and to skip to the next and previous song.

PR #15, opened 2016-11-24
[modules] Show updates for pacman.

Pseudonick47:
Hey there! My friend @paxy97 and I found your bar and really liked it. Here is the module similar to your dnf but for ArchLinux.
I didn't add themes for the module.

### DIFF
--- a/bumblebee/bin/customupdates
+++ b/bumblebee/bin/customupdates
@@ -1,0 +1,21 @@
+#!/usr/bin/bash
+if ! type -P fakeroot >/dev/null; then
+	error 'Cannot find the fakeroot binary.'
+	exit 1
+fi
+
+if [[ -z $CHECKUPDATES_DB ]]; then
+	CHECKUPDATES_DB="${TMPDIR:-/tmp}/checkup-db-${USER}/"
+fi
+
+trap 'rm -f $CHECKUPDATES_DB/db.lck' INT TERM EXIT
+
+DBPath="${DBPath:-/var/lib/pacman/}"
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Great to hear that, thanks a lot!

I've committed slight changes to your module in 8d53b2bedda52e8eb5583f3bb4f70b87f9790587 and 8ceeac84c01c483412669d66ecb95ab0966d7fa1).

Since I don't have ArchLinux installed, I'd be very grateful if you could test the changes to ensure they work.

The changes are:
* I moved the bin/ directory to the root of the project. That way, bumblebee/ remains the directory for all Python modules (I'll also have to move the themes/ directory eventually)

* The pacman module contains a path calculation, which I modified to use os.path.dirname() in order to avoid having to have fixed path lengths in there (the path[:-9]). No big deal, but from experience, I'd say that way it's harder to break in the future :)

Thanks a lot for your contribution!

paxy97:
Yeah, we were looking for a better solution for the second issue, but couldn't find anything. Like I said we're both pretty new to python, so any suggestions and corrections are appreciated :)

Pseudonick47:
Yea, this was pretty hardcoded solution :)
I tested these changes you made and so far they work just fine.

tobi-wan-kenobi:
Great to hear that! Please feel free to also just drop me a message if you have a question regarding any further module you might want to implement. I'm always happy to help, if I can!

Pseudonick47:
Sure!

PR #25, opened 2016-11-30
Update nic.py

hubertst:
Exclude veth* interfaces created by minimesos


tobi-wan-kenobi:
Thanks!

Have to admit: First time I've ever heard of minimesos :-)

I'll have to finally get around and make the list of excluded interfaces a config parameter.

PR #33, opened 2016-12-12
[theme/gruvbox] Remove empty space around the modules

paxy97:
These margins were driving me crazy so I removed them.


tobi-wan-kenobi:
:-) If that slipped in during my recent merge - sorry for that!

PR #44, opened 2017-02-26
[modules/weather] Add rudimentary weather module

paxy97:
It searches for the current ip's location and calls the openweathermap api to get the current weather. I put my own api key for now, I don't need/use it for anything else so if you're okay with it it can stay.

### DIFF
--- a/bumblebee/modules/weather.py
+++ b/bumblebee/modules/weather.py
@@ -1,0 +1,45 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the temperature on the current location based on the ip
+
+Requires the following python packages:
+    * urllib
+
+"""
+
+import bumblebee.input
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks for the module, nice work!

FYI: I've modified the module a bit (e.g. the `urllib` import didn't work for me, also, the unit tests were complaining about missing encoding). It would be great if you could check whether the module still works for you.

paxy97:
It does, maybe it has to do with the version of urllib installed, because most examples I found used urllib2 instead of just urllib, but that one didn't work for me.

tobi-wan-kenobi:
Yeah, looks like the import for urllib works differently for python2 vs. python3. In the end, I simply replaced urllib with requests, which has a consistent interface :-)

PR #47, opened 2017-03-03
new module that displays current kernel version

pierre87:
it uses the output of 'uname -r' prefixed with the linux font-awesome icon


tobi-wan-kenobi:
great, thanks!

PR #48, opened 2017-03-09
Update update module

yashar-sb-sb:
Used shutil.disk_usage(path) instead of os.statvfs(path)
disk_usage is more accurate and also makes code more readable.

### DIFF
--- a/bumblebee/modules/disk.py
+++ b/bumblebee/modules/disk.py
@@ -1,6 +1,7 @@
 # pylint: disable=C0111,R0903
 
 """Shows free diskspace, total diskspace and the percentage of free disk space.
+-
 Parameters:
     * disk.warning: Warning threshold in % of disk space (defaults to 80%)
     * disk.critical: Critical threshold in % of disk space (defaults ot 90%)
--- a/bumblebee/modules/disk.py
+++ b/bumblebee/modules/disk.py
@@ -1,14 +1,13 @@
 # pylint: disable=C0111,R0903
 
 """Shows free diskspace, total diskspace and the percentage of free disk space.
-
 Parameters:
     * disk.warning: Warning threshold in % of disk space (defaults to 80%)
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi and many thanks for the PR!

I agree that using `shutil.disk_usage()`, as per your diff, makes the code more readable, however, as far as I can find out, this method was introduced only in Python 3.3. So either, we'd need to drop support for older Python versions or we'd need to have a fallback (which would complicate the code again).

You have another good point regarding the accuracy of `shutil.disk_usage()` vs. `os.statvfs()`, which, I think comes from the fact that I'm not correctly considering reserved blocks. I will investigate how to improve that.

What's your take on this?

yashar-sb-sb:
I looked at the `shutil.py` file and found how it's implemented.
it uses `os.statvfs` in order to compute `disk_usage`.
I can copy it's code.

yashar-sb-sb:
I copied the code from `shutil.disk_usage`.

tobi-wan-kenobi:
Thanks a lot, this looks very nice!

Just a small note: I've removed the unicode character "f0a0" from the output, I think that was an unintentional change, right?

yashar-sb-sb:
No it was intentional. I should have written a more clear commit message for it.
It's just a disk Icon for aesthetics. 

yashar-sb-sb:
Oh, I've set the Icon in the wrong place.
What is the standard way of setting Icons in `modules`?

yashar-sb-sb:
I looked at the codes.
I noticed that it's theme's job to supply icons.

tobi-wan-kenobi:
Yeah, exactly right. If you want to change the icons, the easiest way would probably be to create a new iconset and use it. (Basically, the icons can be specified as "prefix" and "suffix" variables, and the iconset JSON is simply merged into the main theme JSON).

PR #60, opened 2017-04-01
Catch RequestException when fetching weather

elshize:
## Problem
Whenever internet connection is lost, and _weather_ module is included, the whole status fails. One possible scenario: you wake up your computer from being suspended, and ﻿the whole status fails before the wifi connection could be established.

## Solution
Simply use `try/catch` when retrieving information from web. Therefore, the temperature will not change.

## Follow-up
Even though this fixes the problem, it's not ideal. Say, you don't have connection at the moment of launch, then the temperature will be 0 by default. However, more natural behavior would be to set it to something indicating no information, like `﻿–` or `?`. @tobi-wan-kenobi, how do you feel about this approach?

### DIFF
--- a/bumblebee/modules/weather.py
+++ b/bumblebee/modules/weather.py
@@ -20,6 +20,7 @@ import json
 import time
 try:
     import requests
+    from requests.exceptions import RequestException
 except ImportError:
     pass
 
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks for the PR! I agree, having a special indicator is preferable. I'll set it to "?" for now.

Issue #18, opened 2016-11-25
[modules/nic] Add traffic counters


tobi-wan-kenobi:
Add rx and tx counters to the nic module to see traffic statistics.

meain:
Can't we close this issue?

tobi-wan-kenobi:
you're absolutely right

closed with #71

Issue #27, opened 2016-12-10
[modules] Fixed-length modules


tobi-wan-kenobi:
The length of some modules (e.g. cmus) fluctuates quite a lot depending on the content. It would be nice to have the option to restrict the length by padding/truncating the contents.

meain:
or scrolling text for something like cmus maybe?

tobi-wan-kenobi:
That's a good idea, I'll look into it. Having something like a maximum configured length (per widget?) and a interval-based one-character scrolling should be do-able, I think.

meain:
We can set the max-width per component and give flags for scrollable components inside a widget ( as one might contain buttons and stuff )

PR #71, opened 2017-04-21
Traffic module

meain:
This is simple traffic widget. ( #18 )

TODO: As the speeds change the size of the widget increases and decreases continuously

### DIFF
--- a/bumblebee/modules/traffic.py
+++ b/bumblebee/modules/traffic.py
@@ -8,10 +8,21 @@ import bumblebee.engine
 
 class Module(bumblebee.engine.Module):
     def __init__(self, engine, config):
-        widgets = []
+        widgets = [
+            bumblebee.output.Widget(name="traffic.down"),
+            bumblebee.output.Widget(name="traffic.up"),
+        ]
         super(Module, self).__init__(engine, config, widgets)
         self._exclude = tuple(filter(len, self.parameter("exclude", "lo,virbr,docker,vboxnet,veth").split(",")))
         self._update_widgets(widgets)
+        self._status = None
+
+    def state(self, widget):
+        if widget.name == "traffic.down":
+            return "down"
+        if widget.name == "traffic.up":
+            return "up"
+        return self._status
 
     def update(self, widgets):
         self._update_widgets(widgets)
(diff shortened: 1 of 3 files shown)


tobi-wan-kenobi:
Hi,

thanks for the module!

Separate icons: The easiest approach would be to have the module produce two widgets, one for upload, and one for download (like you already do) and give them different states (this is used in the module `cmus` to have prev/next widgets).
Alternatively, we can think about a way to have arbitrarily insertable glyphs in the theme (if I manage to come up with something that doesn't look horrible in the theme spec).

I hope to get around to having a look at #27 this weekend, which would hopefully enable you to fix your "TODO".

This seems to be a module I'd like to have a careful look at before committing (because it's really cool functionality), so please bear with me if it takes a while to merge the PR.

meain:
module:
```python
        widgets = [
            bumblebee.output.Widget(name="traffic.down"),
            bumblebee.output.Widget(name="traffic.up"),
        ]
```

awesome-fonts:
```json
  "traffic":{
	  "down": { "prefix": "" },
	  "up": { "prefix": "" }
  }
```

Anything else I got to get the icon showing up?

meain:
I now has separate up and down speeds.

tobi-wan-kenobi:
Hello,

I have reviewed you module and have some small suggestions, which I'd like to discuss with you. You can find my proposed modifications here: https://github.com/tobi-wan-kenobi/bumblebee-status/tree/meain-traffic

Basically, the changes are:
- Show statistics per NIC (first widget: name of NIC, second: downlink traffic, third: uplink traffic)
- Use `psutil` instead of `ifconfig`, because this removes a dependency (the tool depends on `psutil` for NIC statistics already)
- Add a min-width and alignment setting to the core framework and the widgets to have a nicer display.

Here's an example how the module looks now:
selection_096

If you like it, I can merge that to master.

meain:
I personally think it is better to keep the up and down icons as prefix for the sake of consistency. Apart from that it looks good.

tobi-wan-kenobi:
Good point, I changed that. Thanks a lot for the contribution, that's a nice module that I'll start using immediately :-)

PR #75, opened 2017-04-24
fix for non-latin symbols in GPM module

graynk:
changes ascii to unicode

### DIFF
--- a/bumblebee/modules/gpmdp.py
+++ b/bumblebee/modules/gpmdp.py
@@ -52,7 +52,13 @@ class Module(bumblebee.engine.Module):
             status = bumblebee.util.execute("gpmdp-remote status")
         except RuntimeError:
             pass
-        self._status = status.decode('utf-8').split("\n")[0].lower()
-        self._tags = info.decode('utf-8').split("\n")[0]
+        try:
+            unicode_status = status.decode('utf-8')
+            unicode_info = info.decode('utf-8')
+        except AttributeError:
+            unicode_status = status
+            unicode_info = info
+        self._status = unicode_status.split("\n")[0].lower()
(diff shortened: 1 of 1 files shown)


graynk:
Be advised that this fix only works for Python 2, and will not compile on Python 3, since Python 3 does not have string.decode() (and does not need this fix in general). As to how to make the code universal for 2 and 3 -- I am not sure.

tobi-wan-kenobi:
Hi,

many thanks for the contribution! I'll review your patch as soon as I have a bit of spare time & try to come up with a Python2/Python3 compatible solution. Might take the weekend, though :)

graynk:
Hi. I think the simple try/except should do the trick. At least it works in the IDE on string with extended symbols for both Python 2 and 3, I don't know how to check if that will work in bumblebee on Python3 though.

tobi-wan-kenobi:
Travis says this fix is OK for both Python 2 and Python 3, the diff looks fine for me too, so I'll just go ahead and merge it. If it breaks, we'll just have to come up with a different solution :)

The more generic (and maybe *slightly* more elegant, but potentially less pythonic) approach would be something like this:

```
if hasattr(status, "decode"):
    status = status.decode("utf-8")
```

tobi-wan-kenobi:
Also, @paxy97 is the original author of the module, so pulling him in, in case he prefers a different solution.

graynk:
Yes, your solution is better, I don't know if it's less pythonic, but at least in Java the use of try-catch _not_ in exceptional situations, where if-else can be used instead is frowned upon. Pretty sure there must be even better way, but it's the only thing that I could think of off the top of my head.

PR #78, opened 2017-05-10
Adds moc module

leggettc13:
More work still needs to be done to sync the repeat and shuffle setting display for the module up properly, but it is good enough for personal use at the moment.

### DIFF
--- a/bumblebee/modules/moc.py
+++ b/bumblebee/modules/moc.py
@@ -1,0 +1,92 @@
+# pylint: disable=C0111,R0903
+# -*- coding: utf-8 -*-
+
+"""Displays information about the current song in cmus.
+
+Requires the following executable:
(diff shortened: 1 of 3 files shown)


tobi-wan-kenobi:
Thanks for the contribution!

From a first glance, the module looks totally fine. Did you close the PR intentionally, to still change some functionality?

leggettc13:
Yeah my bad, I was a bit hasty and decided it wasn't ready. The main problem was that mocp does not output settings for shuffle and repeat, so if you change it from within the app itself it doesn't sync up with the module. Now there's a mocp module that someone else contributed so mine is pretty much pointless now anyway.

tobi-wan-kenobi:
Thanks for the contribution anyhow :-)

PR #83, opened 2017-05-16
Fixed the bug of `ping.interval` raising an exception with any value

Majora320:


tobi-wan-kenobi:
Thanks for the fix! No idea how that slipped by...

Note to self: This calls for a unit test :-)

PR #86, opened 2017-05-26
[modules/github] Add rudimentary GitHub unread notifications module

yvesh:
GitHub Plugin, which shows unread notifications (up to 50 by GitHub API default limit). Refreshes every 5 minutes by default. Requires curl.

Parameters:

`github.token`: Access token for GitHub
`github.interval`: Refresh interval in minutes

### DIFF
--- a/bumblebee/modules/github.py
+++ b/bumblebee/modules/github.py
@@ -1,0 +1,47 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the unread github requests of a user
+
+Requires the following executable:
+    * curl
+
+Parameters:
+    * github.token: GitHub user access token
+    * github.interval: Interval in minutes
+"""
+
+import bumblebee.input
+import bumblebee.output
+import bumblebee.engine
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi and thanks a lot for your contribution, I like the idea of the module a lot (= I'll probably use it myself).

I have a few suggestions that I added in https://github.com/tobi-wan-kenobi/bumblebee-status/tree/yvesh-githubunread. Unfortunately, I was unable to perform "real" testing, as I don't have any unread notifications.

It would be great if you could review my adaptions. If you are happy with them, and they work for you, I can merge them back to master.

yvesh:
@tobi-wan-kenobi just tested your changes, works without issues :-) Thought on using requests, but always used curl in the other bars i am using. But makes way more sense for Python.

Thank you!

> Unfortunately, I was unable to perform "real" testing, as I don't have any unread notifications

Lucky you :)

PR #89, opened 2017-05-26
[Output] Only draw widget when full text is not empty

yvesh:
Currently an small empty widget is also drawn when the output of the command is empty. This PR changes this behavior.


tobi-wan-kenobi:
looks good, thanks!

PR #92, opened 2017-05-30
[modules/publicip] Add module to show current public IP address

RobertSacks:
I added a module that shows your current public IP address. I find it useful when I'm connected to VPNs so I know where I'm reaching the internet from.

It can be added just like any other module by putting `publicip` in the modules list.

P.S. I love using this status bar in i3; thank you so much @tobi-wan-kenobi!

### DIFF
--- a/bumblebee/modules/publicip.py
+++ b/bumblebee/modules/publicip.py
@@ -1,8 +1,5 @@
 """Displays public IP address
 
-Parameter:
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Hi,

thanks a lot for the module, that's a useful one!

I'm glad you like the status bar :)

PR #93, opened 2017-06-01
[modules/memory] Add parameter to only show used memory

ibrokemypie:
No change to default behaviour, but adds boolean to only display used rather than used, total and percentage.

To only show used  memory:
-p memory.usedonly=1


tobi-wan-kenobi:

PR #94, opened 2017-06-01
Simple bluetooth module with toggling and status display

brunosmmm:
Must have bluez / blueman in order to toggle status, and "dbus-send" utility.


tobi-wan-kenobi:
This looks nice and usable pretty generically already (what with the configurable left-mouse application and the device parameter), thanks!

PR #96, opened 2017-06-02
Pop-up menus!

brunosmmm:
This is just an experimental feature that I have been playing with. Please consult the modified bluetooth module to see how it works. I don't see a need to merge this straight away but I thought it'd be nice to share.

### DIFF
--- a/bumblebee/modules/bluetooth.py
+++ b/bumblebee/modules/bluetooth.py
@@ -14,6 +14,8 @@ import bumblebee.input
 import bumblebee.output
 import bumblebee.engine
 import bumblebee.util
+import bumblebee.popup
+import logging
 
 
 class Module(bumblebee.engine.Module):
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Oh, this is a nice idea, thanks! I need some time to review that in detail, but already now: Thanks a lot! I appreciate creative input like this a lot :)

codingo:
Definitely keen to see this go live - I'd love to not have to use a wifi widget alongside Bumblebee and this would allow the wifi module to be expanded for the same kind of functionality.

tobi-wan-kenobi:
Hi,

as far as I can tell, the code looks good, and most importantly: I don't think it breaks any existing functionality (I made a small change to make the tkinter import optional). So, I'm adding this to master in the hopes that somebody (@codingo) finds it useful :-)

Just one caveat: Please treat the popup functionality as somewhat experimental, as I didn't have time to doo extensive testing. But I think this is a really nice addition that we can hopefully expand upon. I really like the interface you chose.

Many thanks for this really cool PR!

PR #97, opened 2017-06-02
[module/cpu] More useful rounding

ibrokemypie:
psutil.cpu_percent() only outputs to one decimal place anyway, so the trailing 0 is useless.
The prepended 0 is also not important, will only be not 0 at 100% utilization, so why not let it be 100% then and take up one less column otherwise?


tobi-wan-kenobi:
The prepended "0"es were intended to achieve a fixed-size widget. Having said that, using a "min-width" would be more elegant.

tobi-wan-kenobi:
Good suggestions, thanks a lot!

PR #99, opened 2017-06-03
[modules/pulseaduio] Allow user choice for giht click command

ibrokemypie:
This allows using eg. pavucontrol-qt or any other command on right click.

### DIFF
--- a/bumblebee/modules/pulseaudio.py
+++ b/bumblebee/modules/pulseaudio.py
@@ -2,6 +2,10 @@
 
 """Displays volume and mute status of PulseAudio devices.
 
+
+Parameters:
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi and thanks for the PR!

this is something that's already possible - though badly (=not) documented.

For **any** module, you can specify the parameter `<module>.left-click` (or `right-click`, etc.) to have custom on-click actions.

For example:

```
bumblebee-status -m load -p load.left-click=halt
```

Is this functionality sufficient for you? If so, I don't think the PR needs to be merged, right?

ibrokemypie:
Ah, had no idea, that works too, means no need for this in any modules, so this can be closed

PR #100, opened 2017-06-03
[themes/icons/awesome-fonts] Replace "Sensors" icon FA alt

ibrokemypie:
The current thermometer icon isn't actually included in fontawesome, requiring a user to install and find another font that does support it, such as symbola or nota-emoji
Since this is for font-awesome it would make sense to only require font-awesome.


tobi-wan-kenobi:
good catch, many thanks!

PR #102, opened 2017-06-03
[modules/traffic] Add option to only show currently up interfaces

### LINKED ISSUE #98: Only Show "Up" interfaces for [modules/traffic]
Similar to a recently closed issue, only show the currently up interfaces in the bar so as not to clutter it with the up and download of offline devices.

ibrokemypie:
Leaves default functionality aas is, but adds simple boollean for disabling the display of interfaces that are down.
Closes #98

### DIFF
--- a/bumblebee/modules/traffic.py
+++ b/bumblebee/modules/traffic.py
@@ -4,6 +4,7 @@
 
 Parameters:
     * traffic.exclude: Comma-separated list of interface prefixes to exclude (defaults to "lo,virbr,docker,vboxnet,veth")
+    * traffic.onlyup: Whether to oly show traffic for Up interfaces (defaults to False)
 """
 
 import re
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi,

I think this has been addressed in 2d99bdf314676b0b110c71805d4428a6e26319bc (by allowing the user to filter the interfaces to display by states). Does that commit suit your purposes, or would you prefer your solution?

Cheers,

ibrokemypie:
Sorry, didnt see that one!
That does probably solve it, however, correct me if I'm wrong, since there are only two states, up or down, would it not be more efficient just to use a simple bool?
Also, since its just up or down, this commit is a lot simpler.

tobi-wan-kenobi:
Good point. My main concern is backwards-compatibility. Conceivably, interfaces can have more states than up/down (i.e. "up, but no IP address", "up, but no internet connectivity", etc.). So my gut feeling was to implement an extensible filter, but I agree that your approach is much simpler and I might have overdone it :)

ibrokemypie:
Ah yeah, that makes sense, up to you then whichever you think is best.

PR #105, opened 2017-06-04
[modules/github] Use header rather than url for token

### LINKED ISSUE #104: Github module passes token in url
```python
            notifications = requests.get("https://api.github.com/notifications?access_token={}".format(token)).text
```

This doesn't appear to be very secure. Anyone sniffing traffic can get the token from the passed url. I'm not that familiar with the Github api, but i can't imagine this can't be done with a post instead?

ibrokemypie:
Doesnt allow sniffers to get the access token from url.
Closes #104

### DIFF
--- a/bumblebee/modules/github.py
+++ b/bumblebee/modules/github.py
@@ -44,7 +44,7 @@ class Module(bumblebee.engine.Module):
                  self._count = 0
                  return
 
-            notifications = requests.get("https://api.github.com/notifications?access_token={}".format(token)).text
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks for the PR!

I **think** using the token in the URL is OK, since the API endpoint is HTTPS, so both the URL and the headers are encrypted using SSL.

Having said that, I also believe that using the "Authorization" header is the more elegant approach.

PR #106, opened 2017-06-05
[WIP] Pep8 Linting

ibrokemypie:
Going through and linting each file, will update as I go.


tobi-wan-kenobi:
Wow, thanks! That's an effort I appreciate a lot!

PR #111, opened 2017-06-10
[modules/nvidiagpu] Nvidiagpu module

RileyRedpath:
Added nvidiagpu module, which parses nvidia-smi to get the GPU name, temperature and memory usage.

Requirements:
* nvidia-smi


tobi-wan-kenobi:

One small hint: You can use `bumblebee.util.execute()` instead of manually doing a `subprocess` call, as this utility function already performs error handling and the UTF-8 conversion.

Cheers!

PR #115, opened 2017-06-14
[Modules/getcrypto] Get crypto prices module 

Ryunaq:
Extremely niche module that will probably never be used for getting the price of BTC, ETH and/or LTC from Kraken.

Btw, this is my first pull request ever so idk if I should do it like I did, anyways, thanks for the awesome bar!

### DIFF
--- a/bumblebee/modules/getcrypto.py
+++ b/bumblebee/modules/getcrypto.py
@@ -6,7 +6,7 @@ Requires the following python packages:
     * requests
 
 Parameters:
-    * getcrypto.interval: Interval in seconds for updating the price, default is 60, less than that will get your IP banned.
+    * getcrypto.interval: Interval in seconds for updating the price, default is 120, less than that will probably get your IP banned.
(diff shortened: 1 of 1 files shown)


Ryunaq:
So as I said, I don't know much about pull requests yet but it seems the reason travis is failing the check is because of the module `test_modules.py`, so I'm not sure if I should try to "fix" that.

tobi-wan-kenobi:
Thanks for the contribution! As far as I can see, the PR looks good. Don't worry about Travis, he likes to complain :) Usually, I fix the failing tests while merging.

If I encounter any troubles, I might come back and ask a few questions, but I think it should work out fine. I will need some time to review the changes, but hopefully, I can merge your changes some time this week.

Ryunaq:
Yep, you were right, changed `_curprice=" "` and that fixed the Travis thing. Hope it works out fine too!

tobi-wan-kenobi:
Looks good, thanks a lot!

PR #121, opened 2017-06-26
Add error handling and the ability to change currencies for the getcrypto module 

quartje:
This pull request adds two things to the getcrypto module: It adds error handling, since Bumblebee would crash after my resuming my laptop. Secondly, it adds the possibility to show the price in other currency than USD, by adding the option getcrypto.getcur. This defaults to USD.


codingo:
I've pulled this into my own project and can verify that it's working as expected however it does add the requests library as a dependency for this module. Given the breadth of requests it seems unlikely that others wouldn't already have it installed (and is already required for the weather module also).

tobi-wan-kenobi:
Many thanks for the PR!

Since @codingo approved the changes (and I had a quick look at the diff and think it looks fine), I'll go ahead and merge to master.

tobi-wan-kenobi:
And as a side remark: @codingo, as far as I can see, the module relied on `requests` already before this PR :-)

PR #122, opened 2017-07-06
github module: move the requests.get inside the try/except

fredj:
otherwise the bar is terminated on network error, also set the output to 'n/a' instead of '0' if the GitHub token is not present


tobi-wan-kenobi:

PR #123, opened 2017-07-07
[module/github] Use json function from requests and 'filter' instead of a loop

fredj:


tobi-wan-kenobi:
I like the change :)

fredj:
So satisfying to remove lines of code :-)

PR #125, opened 2017-07-07
[module/memory] Fix parameter prefix in doc

fredj:


tobi-wan-kenobi:
Ah, that's an embarrassing one :P Thanks for the fix!

PR #128, opened 2017-07-07
Update module dependencies in readme

ibrokemypie:
A few more now depend on requests, which were not added accordingly.


tobi-wan-kenobi:
Hah, overlooked those. Updated it in the Wiki, but not the README.

Many thanks!

PR #130, opened 2017-07-08
Add bumblebee.util.asbool function

fredj:
Harmonize the boolean parameter value. Now `t`, `true`, `y`, `yes`, `on`, `1` are considered truthy and
everything else falsy.


tobi-wan-kenobi:
Oh, that's a nice solution. This has been bugging me for a while already :-)

+ Thanks for the fact that all your PR so far pass all tests, makes them really easy to merge!

PR #132, opened 2017-07-08
[module/{load|cpu}] Use the thresholds from the parameters

### LINKED ISSUE #131: warning and critical parameters in cpu module
[According to the doc](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/bumblebee/modules/cpu.py#L5-L7) it's possible to change the warning and critical threshold but the values are not used in the code. 

How does it work?

fredj:
fixes #131

### DIFF
--- a/bumblebee/modules/cpu.py
+++ b/bumblebee/modules/cpu.py
@@ -21,6 +21,8 @@ class Module(bumblebee.engine.Module):
         widget = bumblebee.output.Widget(full_text=self.utilization)
         widget.set("theme.minwidth", "99.9%")
         super(Module, self).__init__(engine, config, widget)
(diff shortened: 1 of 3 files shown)


tobi-wan-kenobi:
Hi,

this change shouldn't be necessary, warning/critical threshold should be available "automagically", since they are such a frequent feature.

If you look at `engine.py` around line 63 (`def threshold_state()`), the values provided to that method are actually only fallback values in case the parameters `critical` or `warning` are not defined.

Is that what you intended in the PR, or am I misunderstanding something?

As a last note: I agree that this behaviour is not obvious, and maybe refactoring `threshold_state()` to be more expressive would make sense.

fredj:
Hi,

I understand now ! thanks for the explanation.

PR #133, opened 2017-07-08
[module/battery] Open gnome-power-statistics on left click

fredj:


tobi-wan-kenobi:
That's a sensible default behaviour, thanks!

PR #139, opened 2017-07-12
[modules/github] Use GITHUB_TOKEN environment variable if present

fredj:
Tested with `export GITHUB_TOKEN=xxx` in my `~/.profile`


tobi-wan-kenobi:
Thanks for the PR!

If that's fine with you, I'd like to close it because it's been replaced with the changes in ab989d8. Is that OK?

PR #141, opened 2017-07-14
[engine] Use RawConfigParser instead of SafeConfigParser

### LINKED ISSUE #140: percent sign in config file
It's not possible to have % in the config file, for example:
```
[module-parameters]
datetime.format = %d/%m/%Y %H:%M
```

fredj:
To allow percent signs in the config file

fixes #140


tobi-wan-kenobi:
Looks like a good change for me, thanks!

PR #143, opened 2017-07-15
[module/memory] Add new format string parameter

fredj:
I was about to add a new `memory.percentonly` parameter but a string format should be much more flexible.


tobi-wan-kenobi:
Looks good to me. **Many** thanks for also adding a test!

PR #146, opened 2017-07-23
[module/batterytimeleft] Add basic battery time left module

yvesh:
Adds a basic module showing how much battery time is left (or unlimited if you are on AC)..

I am a bit unsure if this should be integrated into the existing battery module or not. But the code is quite different.

What do you think @tobi-wan-kenobi?

Requires python power package.

### DIFF
--- a/bumblebee/modules/batterytimeleft.py
+++ b/bumblebee/modules/batterytimeleft.py
@@ -1,0 +1,57 @@
+"""Displays how much battery time is left.
+
+Requires the following library:
+    * power
+"""
+
+import os
+import glob
+import functools
+
+import bumblebee.input
+import bumblebee.output
+import bumblebee.engine
+import bumblebee.util
+
+try:
+    import power
+except ImportError:
+    pass
+
+class Module(bumblebee.engine.Module):
+    def __init__(self, engine, config):
+        super(Module, self).__init__(engine, config,
+                                     bumblebee.output.Widget(full_text=self.estimate)
+                                     )
+        self._estimate = "n/a"
+        engine.input.register_callback(self, button=bumblebee.input.LEFT_MOUSE,
+            cmd="gnome-power-statistics")
+        immediate_update = functools.partial(self.update)
+        engine.input.register_callback(self, button=bumblebee.input.RIGHT_MOUSE,
+                                       cmd=immediate_update)
+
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Nice, thanks a lot!

I believe it would be good to integrate it into the existing module, as it's fundamentally about the same information. Maybe it would make sense to show it in a second, optional, widget? In that case, if the "power" package is not available, the second widget could show "n/a" instead of terminating the bar, maybe.

How does that sound to you, @yvesh ?

If it's fine for you, I can try to do the changes next weekend (this week's more busy than usual, unfortunately).

Cheers,

tobi-wan

tobi-wan-kenobi:
Short update: I did a quick implementation and pushed it into the branch pr146-battery-remaining. @yvesh , if you want, you can have a look and let me know how you like it!

I left out the immediate update, as I didn't find a way to make it work (since the battery module has a variable number of widgets).

yvesh:
Looks great @tobi-wan-kenobi! Going to do some tests tomorrow. Imo it could look better if we just put it next to the battery percentage, maybe in squares, like (5:30 h remaining). E.g. skip the icons totally. What do you think? Going to play a bit around and make some screenshots.

tobi-wan-kenobi:
Good point, I'll try to implement that as soon as possible, then let's have a look. I think I'd prefer the solution you suggested.

yvesh:
Okay played a bit around, my favorite is within the battery. 

Some screenshots for choosing:





Estimate a bit better formatted:

```python
self._estimate = str(datetime.timedelta(minutes=estimate)).split('.', 2)[0] + ' h'
```

P.S.: Imo this setting should be enabled by default, because at least for me it's really helpful and one of the few features missing ;-) 

Thank you for all your work @tobi-wan-kenobi!

tobi-wan-kenobi:
@yvesh thanks a lot! I would appreciate your input on the updated branch a lot. I have:

* included the remaining time directly into the widget
* hidden the remaining time if on charge (unless the module displays time until full charge then?)
* shortened the output to not show seconds if hours+minutes are available

tobi-wan-kenobi:
merged to master

yvesh:
Thank you @tobi-wan-kenobi! Sorry for not testing earlier, but looks great so far!

PR #147, opened 2017-07-24
Add click events to Spotify module

rccavalcanti:
Added support for click events on Spotify.

Left click: previous song
Right click: next song
Middle click: play/pause.

It uses dbus-send so no additional dependencies AFAIK.

Thanks @tobi-wan-kenobi for the great project.


tobi-wan-kenobi:
Nice! Thanks for the PR!

PR #148, opened 2017-07-25
[module/github] Add pagination support

fredj:
The GitHub API returns 50 items per page.


tobi-wan-kenobi:
Nice change! I particularly like the use of `requests.Session`

Thanks!

PR #149, opened 2017-07-27
Locale for datetime module

vidyu:
Allows usage of strings like %x, %X for the datetime format.


tobi-wan-kenobi:
:+1: I like, thanks for the PR!

PR #150, opened 2017-07-30
[module/battery] Simplify remaining time computation

fredj:
The `power` module already return a special value on AC: `power.common.TIME_REMAINING_UNLIMITED`


tobi-wan-kenobi:
Nice, thanks!

PR #151, opened 2017-07-30
disk module: added remaining space option + bug fix

Kymon:
Enhancement of disk module which shows remaining space on disk.
Turning it on/off with parameter disk.showRemain="yes/no" (default yes).

Fixed bug caused by disk.showPercent="no".

### DIFF
--- a/bumblebee/modules/disk.py
+++ b/bumblebee/modules/disk.py
@@ -7,9 +7,7 @@ Parameters:
     * disk.critical: Critical threshold in % of disk space (defaults ot 90%)
     * disk.path: Path to calculate disk usage from (defaults to /)
     * disk.open: Which application / file manager to launch (default xdg-open)
-    * disk.showUsed: Show used space (defaults to yes)
-    * disk.showSize: Show total size (defaults to yes)
-    * disk.showPercent: Show usage percentage (defaults to yes)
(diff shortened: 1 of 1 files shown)


fredj:
And what about adding a new `format` option like it was done for the memory module in #143?

Kymon:
That would be neat. I will change this soon.

Kymon:
I guess everything is done.

tobi-wan-kenobi:
That looks great, thanks!

One small thing, though: I'd love it if we could keep backwards compatibility by setting the default format string depending on whether any of the "show" flags are present (we can mark them as "deprecated", though). Does that sound fine for you?

I can do the change myself, but will probably only have time on the weekend, I'm pretty busy at work, right now. :(

tobi-wan-kenobi:
@Kymon what do you think about the proposed patch?

It makes the code less beautiful (thanks for the nice cleanup, by the way!), but it should be backwards-compatible now.

fredj:
@tobi-wan-kenobi you're patch looks good. I'm not using the battery module so I'm not the best person to ask ...

Kymon:
@tobi-wan-kenobi it's fine.

tobi-wan-kenobi:
thanks a lot for your feedback!

PR #154, opened 2017-08-04
[module/spotify] Add new format string parameter

fredj:
Same idea as #143 but for the spotify module


tobi-wan-kenobi:
Looks good, thanks! I'll keep the "format" string convention in mind for future modules :)

PR #156, opened 2017-08-06
CPU icon and datetime locale

nayaverdier:
I'm not sure why the CPU icon was changed in df6e323 but it stopped working for me. This PR reverts that and also adds functionality to override the locale in datetime.


tobi-wan-kenobi:
looks good, thanks!

I will make a small change reg. handling the case when no locale is provided (to reduce code).

PR #160, opened 2017-08-17
[Modules/mpd] Add MPD host parameter

ibrokemypie:
MPC uses localhost by default, but some configurations require otherwise.
Default behaviour is unchanged.


tobi-wan-kenobi:
Looks good to me, thanks for the change!

PR #162, opened 2017-08-21
fix: exception occurred when screen is locked

zetxx:


tobi-wan-kenobi:
look good to me, thanks!

PR #169, opened 2017-09-07
[modules/fx]

AntouanK:
Add an fx module to track GBP/EUR and GBP/USD.

I made this just for personal use, but I thought someone else might want to see currency pairs as well.
I'm not a python dev, so the code is probably horrible.
If someone wants to clean it up or add a way to parameterise the currency pairs, go ahead.

@tobi-wan-kenobi  Thanks for all your work on bumblebee and it's modules.

### DIFF
--- a/bumblebee/modules/fx.py
+++ b/bumblebee/modules/fx.py
@@ -1,0 +1,51 @@
+# -*- coding: UTF-8 -*-
+# pylint: disable=C0111,R0903
+
+import bumblebee.input
+import bumblebee.output
+import bumblebee.engine
+import json
+import time
+try:
+    import requests
+    from requests.exceptions import RequestException
+except ImportError:
+    pass
+
+class Module(bumblebee.engine.Module):
+    def __init__(self, engine, config):
+        super(Module, self).__init__(engine, config,
+            bumblebee.output.Widget(full_text=self.price)
+        )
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi,

that's a nice module, many thanks for the PR!

I will merge it for sure, but before I do so, some questions:

* Would it be OK for you if I rename it to "currency"? I think it's easier to guess its purpose then.
* If that's fine with you, I'd have a go at changing the module so that it creates one widget per currency pair, that would make it more extensible, I believe.

If that's fine with you, I'll try to get this done during the weekend, at latest.

Thanks a lot!

AntouanK:
Feel free to change it in any way you want.
Like I said, I only did an attempt because I find it useful to check those two pairs every day.
I know there's one for crypto currencies as well ( I use that too ) so maybe merge them to have one module for all currency pairs?

AntouanK:
Thank you @tobi-wan-kenobi

tobi-wan-kenobi:
Hi,

I've now added the option to configure base and symbols. Doing that, I change the output format a bit, I hope you are OK with it.

What's still open (= would be a nice addition) is a way to configure the output format, but since part of it is variable (depending on the number of destination symbols), I'm a bit unsure on how to make that configurable in a nice manner.

PR #170, opened 2017-09-09
Removed returning n/a

batman-nair:
Better not to show.


tobi-wan-kenobi:
Thanks for the PR! I agree, this change makes the status bar a bit more compact.

PR #171, opened 2017-09-09
Added parameter showname to hide network interface

batman-nair:
Becomes a needless info for personal laptop usages where only one interface is used.


tobi-wan-kenobi:
Thanks for the PR! I've merged it to master with a few small modifications (there's a helper function in `bumblebee.util` to convert parameters to boolean, so that we have a consistent boolean representation).

PR #174, opened 2017-09-15
[modules/title]

UltimatePancake:
A little feature I've always liked so as to not deal with title-bars.

Parameters for max length and placeholder text.

Truncated:
sep15_101705_s

Not-truncated:
sep15_101643_s

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.5"
   - "3.6"
 install:
+  - pip install i3ipc
   - pip install psutil
   - pip install netifaces
   - pip install -U coverage==4.3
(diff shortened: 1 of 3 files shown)


tobi-wan-kenobi:
Ooooohhh, *that* is a nice one, thanks!

tobi-wan-kenobi:
A few notes/suggestions:

1. I replaced `textwrap` with a custom implementation for compatibility with older Pythons, hope that is fine.
2. In the same direction, how about using `bumblebee.output.scrollable`, like `cmus` does? Would you find that suitable, or too distracting, due to the constant movement?
3. Unicode characters break the module right now, but I'm too tired to investigate right now :-)

PR #175, opened 2017-09-15
[module/title] Scrolling title

UltimatePancake:
> I replaced textwrap with a custom implementation for compatibility with older Pythons, hope that is fine.

Good one, didn't think of that, sorry (should really get into the habit of using travis).

> In the same direction, how about using bumblebee.output.scrollable, like cmus does? Would you find that suitable, or too distracting, due to the constant movement?

Why not both? Options are always nice. ;)

> Unicode characters break the module right now, but I'm too tired to investigate right now :-)

Will look into that when I catch a bit more time.


tobi-wan-kenobi:
*like*

reg. unicode: will look into that, too, as per #176 ;-(

PR #178, opened 2017-09-17
BUG: github module count breaks in python3

chrismullins:
In python2, `filter()` returned a list, but in python3 it returns an iterator. So
we wrap this in a `list()` so that it works in both.  We also want to count the
unread notifications, so this should be reflected in the code.


tobi-wan-kenobi:
Nice one, thanks a lot!

PR #183, opened 2017-09-29
modules/sensors: Try intel scaling current frequency

ibrokemypie:
On intel pstate drivers only the normal frequency is written to /proc/cpuinfo, not accounting for scaling/turbo.
This should fix that, though it needs testing on other computers (AMD, intel without pstate) because I am not
certain that the file only exists on Intel pstate.


tobi-wan-kenobi:
Looks good to me, let's see what others report :-)

PR #184, opened 2017-09-29
[modules/weather] Show city in weather module

paxy97:
Also show the current city in the weather module

https://i.imgur.com/ghKgPPg.png


tobi-wan-kenobi:
Great, thanks!

PR #186, opened 2017-10-07
nvidiagpu.py: remove invalid shebang

avindra:
This file has no actionable code when executed directly from a shell. It contains only a `class` as it is a bumblebee module.

so the shebang should be removed.


tobi-wan-kenobi:
Good catch, thanks!

PR #187, opened 2017-10-11
[modules/rotation] new module to loop through orientations

cacyss0807:
New module to change the orientation of a specific display. 
* Instead of having a static default orientation to change to when the widget is toggled, this module cycles through all possible orientations.
* loosely based on the `xrandr` module.


tobi-wan-kenobi:
Nice, many thanks for the PR!

I'll test it at work, where I have a multi-monitor setup, but I'll go ahead and merge it in the meantime, as the code looks good to me.

PR #192, opened 2017-10-17
[modules/current] Use markets.ft.com instead of fixer.io

### LINKED ISSUE #190: Currency module
@tobi-wan-kenobi Regarding the currency module.
So far, I used the "fixer.io" endpoint. The good thing is that it can give you many pairs in one call, but unfortunately it only updates once a day.

I wanted to change it to use the internal FT endpoint.
_example for GBP/USD_
`https://markets.ft.com/data/currencies/ajax/conversion?amount=100&baseCurrency=GBP&comparison=USD`
But, apart from changing the way to access the value from the JSON response ( that's simple, just change the keys ) the difference is that here we need one call per pair.
But, this one updates every few minutes, so it's much more useful.

My python knowledge is non existent, so can you give it a try to change it?
Somehow can we recognise the arguments, and add 1 call for each argument in the "loop"?

I guess the separator will become optional as well, depending on whether you have more than 1 arguments.

fredj:
Fixes #190 

I didn't use the `currency` parameter because only the rate is displayed (not sure this is correct).

Comments welcome!

### DIFF
--- a/bumblebee/modules/currency.py
+++ b/bumblebee/modules/currency.py
@@ -14,7 +14,7 @@ Parameters:
                              the base symbol and the rate list
     * currency.destinationdelimiter: Delimiter used for separating individual rates (defaults to "|")
 
-Note: source and destination names right now must correspond to the names used by the API of http://fixer.io
+Note: source and destination names right now must correspond to the names used by the API of https://markets.ft.com
 """
 
 import bumblebee.input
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good to me, thanks!

I'll just merge it to master, as it passes all tests, and we'll see if there's any reports coming in about broken functionality ;)

AntouanK:
Thank you @fredj . I'll let you know if I see any problems with it.
I'm only worried about the API call interval. Let's see if FT blocks it after a while.

PR #198, opened 2017-10-25
Minor modules modifications

cacyss0807:
Minor update to `rotation` and `cpu` modules

* `cpu`: added format parameter.
* `rotation`: re-create a widget on the bar per connected display only during the refresh of i3-bar


tobi-wan-kenobi:
Thanks a lot for the PR! Looks good to me from the diff, and Travis also doesn't complain - so in to master it goes :)

PR #199, opened 2017-10-27
[modules/title] fixed runtime exception

cacyss0807:
From i3ipc the find_focused().name can return  a None instead of a string, this will casue a runtime exception


tobi-wan-kenobi:
Thanks a lot for the PR, makes a lot of sense, so I'll just go ahead and merge it to master!

PR #200, opened 2017-10-27
Add new hipchat module

fredj:
**work in progress**

Displays the unread messages count for an HipChat user

I still want to add an icon and open the app on left click


tobi-wan-kenobi:
Nice one, thanks! As soon as there's an icon, I'll add a screenshot to the list of apps, if you can provide one (I'm not using HipChat anymore, myself).

fredj:
Thanks @tobi-wan-kenobi !

PR #204, opened 2017-11-13
[modules] added new shortcut module

cacyss0807:
This modules makes it very easy to create shortcuts as widgets, for which the user can define the command to be executed when left clicking on it. It supports single or multiple shortcuts

### DIFF
--- a/bumblebee/modules/shortcut.py
+++ b/bumblebee/modules/shortcut.py
@@ -1,0 +1,70 @@
+# pylint: disable=C0112,R0903
+
+"""Shows a widget per user-defined shortcut and allows to define the behaviour
+when clicking on it.
+
+For more than one shortcut, the commands and labels are strings separated by
+a demiliter (; semicolon by default).
+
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good, thanks a lot for the PR!

I'll go ahead and merge it, as it is green in Travis.

Just one idea: My first intuition was that it would be nicer if it would be possible to configure labels and actions together, instead of as two separate variables, each being delimited. The problem, of course, is that then you'd suddenly have *two* reserved delimiters. Maybe one solution would be to add array support to the parameters, like: `-p cmd=cmd;label cmd=cmd2;label2 cmd=cmd3:label3`

But frankly, that also looks like a workaround, so I don't have any good ideas right now :P

cacyss0807:
I agree we could find a better way. The array support seems like an interesting idea. Any parameter that has to be passed in as tuples would be appended to a parameter list, but then the order would matter (i.e. command before label).

PR #205, opened 2017-11-13
[modules/uptime] Add uptime module

ccoors:
I was surprised this didn't exist yet.

Note that I'm not a python developer, so I assume the code is not perfect. I'd be happy to change anything. I tried to add formatting (like %H:%M:%S), but I didn't find an easy way to do it.

### DIFF
--- a/bumblebee/modules/uptime.py
+++ b/bumblebee/modules/uptime.py
@@ -1,0 +1,26 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the system uptime."""
+
+import bumblebee.input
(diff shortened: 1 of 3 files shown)


tobi-wan-kenobi:
Hi and thanks a lot for the PR! Yeah, surprisingly, there was no `uptime` module until now, so great that you provide it!

I think the formatting is fine as it is, as Python will take care of the correct string representation, hopefully :)

I'll go ahead and merge it and make sure the tests run through (I think Travis is just missing a module)

PR #207, opened 2017-11-18
Use absolute_import in uptime module

### LINKED ISSUE #206: uptime module: import error
There's an `ImportError` from the uptime module: 
```
  File "/home/fredj/bin/bumblebee-status", line 73, in <module>
    main()
  File "/home/fredj/bin/bumblebee-status", line 45, in main
    inp=inp,
  File "/home/fredj/code/bumblebee-status/bumblebee/engine.py", line 117, in __init__
    self._aliases = self._read_aliases()
  File "/home/fredj/code/bumblebee-status/bumblebee/engine.py", line 194, in _read_aliases
    mod = importlib.import_module("bumblebee.modules.{}".format(module["name"]))
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/fredj/code/bumblebee-status/bumblebee/modules/uptime.py", line 9, in <module>
    from datetime import timedelta
ImportError: cannot import name timedelta
```
The reason is that python tries to import `timedelta` from `datetime` in the same directory and not from the standard lib.

I'm using python 2.7.12

fredj:
fixes #206 

Only tested with python 2.7.12

### DIFF
--- a/bumblebee/modules/uptime.py
+++ b/bumblebee/modules/uptime.py
@@ -2,6 +2,10 @@
 
 """Displays the system uptime."""
 
+# Use absolute_import because there's already a datatime module
+# in the same directory
+from __future__ import absolute_import
+
 import bumblebee.input
 import bumblebee.output
 import bumblebee.engine


tobi-wan-kenobi:
Nice, thanks! I solved that by renaming `datetime` to `date-time`, but your solution is much less likely to break something.

PR #208, opened 2017-11-25
[modules] Add zpool module

adam-dej:
This PR adds a module showing zpool stats and zpool iostats for ZFS users.


tobi-wan-kenobi:
Wow, that's a great PR, thanks a lot!

PR #217, opened 2018-01-04
Remove stock module

avindra:
Per Yahoo, this service is now discontinued. The CSV file merely returns
an error message:

 > It has come to our attention that this service is being used in
   violation of the Yahoo Terms of Service. As such, the service is
   being discontinued. For all future markets and equities data
   research, please refer to finance.yahoo.com.

Or, try a direct link:

http://download.finance.yahoo.com/d/quotes.csv


tobi-wan-kenobi:
Damn :(

At a quick glance, I cannot find any other free API, also Google Finance gives me error messages.

There's https://www.alphavantage.co/, but I want to investigate that a bit further first.

Having said that, I'm not sure removing the module is the best option, as we'd "just" need to find a different API provider.

Maybe having an option to mark it "out of service" or something similar and show a corresponding errors message. Otherwise, existing configurations would break.

What do you think?

avindra:
Yeah maybe having an "out of service" message would be the way forward. Institutions that hold financial data tend to be pretty uptight about how their services are used, so it's likely that  this would happen quite often.

tobi-wan-kenobi:
Tracking those issues in #219 and #220. Closing the PR in favour of those.

Thanks a lot!

PR #225, opened 2018-01-30
[modules] add taskwarrior module

chdorb:
This PR is about a new module, taskwarrior, which displays the number of pending tasks in the [TaskWarrior](https://taskwarrior.org/) tool.


tobi-wan-kenobi:
Very nice, thanks for the PR!

I see that for Python 3, Travis is still failing, but at first glance, this looks like an issue in the task warrior library (???).

I'll try to find some time during the weekend to dig into that. Please stay tuned :)

chdorb:
I haven't made a deep investigation, but for Python 3, the taskw package make use of the taskwarrior tool, and the task command is not available in the test environment.
See https://github.com/ralphbean/taskw/blob/f813ca309df01775c4a041fb47827298224dd9e1/taskw/warrior.py#L3

tobi-wan-kenobi:
Sounds good, I'll go ahead and merge the PR, then try to get Travis to include the binary :)

PR #226, opened 2018-01-31
[modules/brightness] Fix return format

caianrais:
Hi.

I've noticed this a while ago and decided to send a pull request. It's a minor issue that bothered me a little bit. It worked in my environment, so I think it's okay.

**What:** Fixes the return format in `brightness` module.
**Why:** To remove the initial zero in the brightness indicator when below hundred.

**Before:**

ss__01-31__06-41-06__309x54
ss__01-31__06-42-10__309x55

**After:**

ss__01-31__07-16-50__306x49
ss__01-31__07-17-00__313x52

Thank you for all the effort put in this project!

### DIFF
--- a/bumblebee/modules/brightness.py
+++ b/bumblebee/modules/brightness.py
@@ -29,7 +29,7 @@ class Module(bumblebee.engine.Module):
 
     def brightness(self, widget):
         if isinstance(self._brightness, float):
-            return "{:03.0f}%".format(self._brightness)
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Yeah, that's an odd one, I personally prefer to have the icons to have a static size, and the easiest way to achieve that was 0-padding :)

Having said that, I'll accept your PR gladly, as there are better ways to do that in the themes by now.

Thanks a lot!

PR #236, opened 2018-03-17
added greyish powerline theme

F00dwarrior:
added another theme which is normally complete grey & black, unless one field is in any special state


tobi-wan-kenobi:
Cool, thanks! Always nice to see new themes!

PR #241, opened 2018-04-01
[modules/traffic] Add parameter to hide down networks

TheEdgeOfRage:
Add a optional parameter to the traffic module to hide networks that are inactive (have no id address). It became frustrating to have it take up a whole screen width of space when I have three interfaces with only one active.


tobi-wan-kenobi:
Nice idea, thanks!

PR #243, opened 2018-04-11
[module/nic] Added format parameter to control the widget output

kunalshetye:


tobi-wan-kenobi:
Nice addition, thanks for the contribution!

Issue #250, opened 2018-04-20
module:[mpd] is not hide when mpd is off


blackjezreel:
when i play music with mpd everything is nice, mpd module is showing in bar.
i just want to know whether it can be hide when the music is off just like cmus and spotify

big thanks

tobi-wan-kenobi:
PR merged - thanks!

PR #251, opened 2018-04-20
Added a hidden method to hide mpd widget when no music is playing

kunalshetye:
Fix for https://github.com/tobi-wan-kenobi/bumblebee-status/issues/250


tobi-wan-kenobi:
Nice, mpd was definitely missing that one! Thanks a lot!

PR #255, opened 2018-05-06
Can force status bar to update early by sending a USR1 signal to the …

WORD559:
…process


tobi-wan-kenobi:
Diff looks good, I particularly like the refactoring. Thanks for the PR!

PR #256, opened 2018-05-06
Adding gruvbox "light" theme

freed00m:
Hi,

I would like to have a light version of the gruvbox theme, I just switched the foreground and background colors in the cycle.

You can see here how it looks. 

2018-05-07-013009_1920x1080_scrot


tobi-wan-kenobi:
Very nice, thanks!

PR #258, opened 2018-05-12
[module] - adding caps/num/scrl indicator widget; pacman to sum the digits

freed00m:
Hi,

My problem is that on my keyboard there are no LEDs that would signal me if Numlock or Caplock are enabled leading me sometimes in a full Rage cause vim stars to jump around.

It uses the "xset q" command like does the caffeine widget.

Some modularity is there too, like changing the color of indicating or any caps "Numlock = nUmLoCk".
2018-05-12-194435_1920x1080_scrot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ ENV/
 
 # Rope project settings
 .ropeproject
+
+# Visual studio project files
+.vscode/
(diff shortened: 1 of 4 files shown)


freed00m:
I don't know about the codeclimate thing, it works. 
And sorry I didn't have wrote a unit test. I just discovered that you wrote a wiki about how to contribute and I did not read. :|

freed00m:
Also, enabling pacman module to view single digit of all updates.

tobi-wan-kenobi:
Thanks a lot, that looks very useful! (my keyboard doesn't have those indicators, either)

Reg. code climate: Don't worry, it's just *very* picky. I'm still trying to find time to figure out how to tweak the metric sensitivity (i.e. not allowing functions to have more than 25 lines seems ... excessive)

Reg. tests: No problem, those are optional :) I'd like to get the test coverage up eventually, but there's also a generic module test that covers *some* aspects of a module anyhow.

Finally: I'll update the pacman module to use `bumblebee.util.asbool()` to "cast" the parameter to bool, for consistency.

PR #265, opened 2018-05-31
Add proper test to check is locale defined. 

4shadoww:
locale.getdefaultlocale() returns (None, None) if locale isn't denifed. So current check isn't enough and fallback locale won't be triggered.


tobi-wan-kenobi:
Hi,

ah, I wasn't aware of that. Many thanks for the PR!

PR #270, opened 2018-06-09
[module/amixer] Add on/off mute check

susmus:
Hi,

When the volume is >0%, but the channel is muted, the amixer module doesn't detect this and still shows Master as unmuted on the statusbar. Example: `Mono: Playback 56 [44%] [-53.25dB] [off]`

Is this intentional? If not, do you think we could add a check for this?


tobi-wan-kenobi:
Hi,

that change makes a lot of sense to me. I think you found a clear-cut bug, thanks for the PR!

PR #274, opened 2018-06-30
fix todo module path expansion

abdoulayeYATERA:
Hi, i found that path expansion was not working in this module.
Just add os.path.expanduser method to handle it.


tobi-wan-kenobi:
Hi,

good catch, many thanks for the PR!

abdoulayeYATERA:
Hey Tobi ! You'r welcome!

Issue #275, opened 2018-07-30
Wrong temperature


eknoes:
#### Summary
Affected module: sensors

Description:
The wrong temperature is shown. I work outside at a temperature >30°C, but the sensors module constantly reports 20°C in my laptop. The module relies on the output of `/sys/class/thermal/thermal_zone0/temp` which can be the right one, but according to [unix.stackexchange](https://unix.stackexchange.com/questions/304845/discrepancy-between-number-of-cores-and-thermal-zones-in-sys-class-thermal) can also be the wrong one.

I will fix it in the next days and do a PR, but am unsure if it is better to 
* iterate through all thermal zones on initialization to find the right type as described on SE
* reverse the order of determining temperature + fallback, so the sensors command will be used if available, otherwise `/sys/.../temp` is read

Maybe a combination of both.

tobi-wan-kenobi:
Fixed in PR #276

PR #276, opened 2018-07-30
[modules/sensors] Prefer sensors command if available, solves #275

eknoes:
As explained in #275, sometimes the sensors module shows a wrong temperature because it simply reads from `/sys/class/thermal/thermal_zone0/temp`. A right way would be to iterate through all thermal zones, determine the valid ones and show the highest temperature. As a quick but not ugly fix, the order how the temperature is retrieved is changed.

Before, it looked up thermal_zone and if it is not available, relies on the `sensors -u` command. Now, it checks on initialization if `sensors -u` is available, if so, it is used, otherwise the thermal_zone is checked.


tobi-wan-kenobi:
Thanks for both the bug report and the PR!

Looks totally fine for me -> will merge.

Thanks for the contribution, much appreciated!

PR #277, opened 2018-08-07
Respect autostart parameter.

mijoharas:
So, a while ago my pulseaudio stopped working (trying to connect with pavucontrol or anything it hangs and says it can't connect). I only just got around to trying to figure out what it was.

So it seems it's due to bumblebee and this specific line in the pulseaudio module. I noticed the auto-start variable so I made sure that this line is only executed if the variable is set.

So I don't _really_ know pulseaudio, but my system (arch) has pulseaudio started by default when it's needed. I wonder if this `pulseaudio --start` messes with that process somehow ([ref](https://wiki.archlinux.org/index.php/PulseAudio#Running)).

Either way, my system is broken without this line.

### DIFF
--- a/bumblebee/modules/pulseaudio.py
+++ b/bumblebee/modules/pulseaudio.py
@@ -29,7 +29,8 @@ class Module(bumblebee.engine.Module):
             bumblebee.output.Widget(full_text=self.volume)
         )
         try:
-            bumblebee.util.execute("pulseaudio --start")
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Many thanks for the PR! I have had the same issue since recently, but never thought to check whether the status bar is causing it :blush:

tobi-wan-kenobi:
FYI: Released 1.6.2 (+ AUR package update) for that fix.

PR #278, opened 2018-08-08
New media player module and Spotify module enhancement

WORD559:
Created a new module for the DeaDBeeF media player based on the module for Spotify. Media controls are the same as the Spotify module for anyone who is already used to this module.

As part of this new module, I added three parameters that will allow users to rebind their media controls. This is documented in the help text. I have also added these parameters to the Spotify module so that users can rebind their media controls there as well.

### DIFF
--- a/bumblebee/modules/deadbeef.py
+++ b/bumblebee/modules/deadbeef.py
@@ -1,0 +1,72 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the current song being played in DeaDBeeF and
+provides some media control bindings.
+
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thanks a lot for the contribution! Also the changes for spotify look nice! As far as I can see, the changes look good, so I'll just go ahead and commit them.

Just one note for the future: Instead of using `subprocess`, you can also use `bumblebee.util.execute()`, which is a bit higher-level.

WORD559:
Ahh, good to know, I'll use that in future. Thanks for merging.

PR #279, opened 2018-08-16
Left mouse click on datetime module opens gsimplecal

david-perez:
This PR adds support for displaying a mini calendar when left clicking the datetime module.

I simply registered a callback to run [`gsimplecal`](https://github.com/dmedvinsky/gsimplecal).

In my i3 config file I have

```
for_window [class="(?i)gsimplecal"] floating enable, move position mouse, move up $height px
```

where `$height` is the height of my status bar.

The result is the following:

selection_020

`gsimplecal` is very configurable and can show clocks for multiple timezones and close automatically when the window is unfocused.

Regarding the added code in this PR, I'm not sure whether we should register the callback unconditionally since I don't think we should make `gsimplecal` a hard requirement for a module as popular as datetime. However, I also don't see any drawback because if the user does not have `gsimplecal` installed, a left click on the datetime module will simply not do anything.

If everything looks good though, I'll go ahead and update the docs.

### DIFF
--- a/bumblebee/modules/datetime.py
+++ b/bumblebee/modules/datetime.py
@@ -37,6 +37,8 @@ class Module(bumblebee.engine.Module):
         lcl = self.parameter("locale", ".".join(l))
         locale.setlocale(locale.LC_TIME, lcl.split("."))
 
+        engine.input.register_callback(self, button=bumblebee.input.LEFT_MOUSE, cmd="gsimplecal")
+
     def get_time(self, widget):
         enc = locale.getpreferredencoding()
         retval = datetime.datetime.now().strftime(self._fmt)


tobi-wan-kenobi:
Thanks for the PR!

Actually, you can achieve the same thing generically by binding the left-click during the invocation like this:

`bumblebee-status -p date.left-click="gsimplecal"`

However, your PR would make that a default behaviour, which is also nice (I typically prefer sensible defaults over explicit configuration). I'm just not 100% sure how many installations have gsimplecal enabled by default. What's your thought on that?

david-perez:
AFAIK `gsimplecal` is not installed by default in any Linux distro. I didn't know about the generic `.left-click=cmd` solution, which I actually prefer. I now lean towards closing this PR and simply documenting somewhere in the wiki the use of `gsimplecal` with the date module to provide a mini calendar, which is a common widget in popular desktop environments.

tobi-wan-kenobi:
Yeah, that sounds good to me. Thanks for raising it, I think a way to document useful hints like this on a per-module basis would be a great addition to the documentation.

PR #282, opened 2018-08-20
Added playback status indicator to spotify module

WORD559:
Small playback indicators for Spotify module. Either "▍▍" when paused, or "▶" when playing. Can be used by putting `{playbackStatus}` in the spotify.format config.

### DIFF
--- a/bumblebee/modules/spotify.py
+++ b/bumblebee/modules/spotify.py
@@ -7,7 +7,7 @@ Requires the following library:
 
 Parameters:
     * spotify.format:   Format string (defaults to "{artist} - {title}")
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looking good, thanks for the PR!

Just two suggestions for future considerations:
1. Maybe that should go into the default string? Personally, I think this would be useful to have enabled as default
2. If there's sufficient interest, it might make sense to model the spotify module like mpd and cmus (i.e. to have separate widgets for controls and title). That would make it easier to interactively pause/unpause/forward/rewind.

But the PR very nice as it is, so I'll just go ahead and merge it :)

Thanks for the contribution!

PR #283, opened 2018-08-21
feat: add notmuch count module

abdoulayeYATERA:
Displays the result of a notmuch count query
default : unread emails wich path do not contained "Trash" (notmuch count "tag:unread AND NOT path:/.*Trash.*/")

Parameters:
    notmuch_count.query: notmuch count query to show result

Errors:
    if the notmuch query failed, the shown value is  -1

Dependencies:
    notmuch (https://notmuchmail.org/)


tobi-wan-kenobi:
Thanks a lot for the PR and the new module!

Issue #285, opened 2018-08-24
bumblebee-status does not find iwgetid on debian


donfranio:
#### Summary
Affected module: nic

Description:
the nic-module does not show the essid on debian systems.

#### What is the current behaviour and what is the expected behaviour?
It just shows no essid but it should.

Probably the problem is, that the called binary ```iwgetid``` resides in ```/sbin``` on debian-systems which is not in the default ```$PATH``` of a typical system user.

#### How to reproduce
bumblebee-status.conf:
```
[module-parameters]

disk.format = Free: {left} ({percent:05.02f}%)
disk.path = /

nic.exclude = lo,virbr,docker,vboxnet,veth,br,ww
nic.states = ^down
nic.left-click = wicd-gtk -n
```

status_command from i3-config
```
    status_command /home/fscherrer/Software/github/bumblebee-status/bumblebee-status\
      -m spotify nic disk cpu battery datetime pasink\
      -t powerline
```

#### Expected behavior
Show the essid of a connected WiFi

tobi-wan-kenobi:
merged PR #286

PR #286, opened 2018-08-24
add 'which' function to utils and use for 'iwgetid'

donfranio:
Solves Issue #285


tobi-wan-kenobi:
Great addition, thanks a lot! The diff looks good to me, Travis seems to have issues because of networking issues, so I'll just merge the changes as is.

Thanks!

PR #287, opened 2018-08-28
add test for util.which

donfranio:
* and make runtests.sh dash-compatible for debian-like OS

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -3,7 +3,7 @@
 echo "testing with $(python2 -V 2>&1)"
 python2 $(which nosetests) -v --with-coverage --cover-erase tests/
 
-if [ $? == 0 ]; then
+if test $? -eq 0 ; then
 	echo
 
 	echo "testing with $(python3 -V 2>&1)"
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -2,6 +2,7 @@
 
 import mock
 import unittest
+import re
 
 import tests.mocks as mocks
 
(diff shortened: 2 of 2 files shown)


donfranio:
I am playing around finding my way to testing in python ... hope you like it.

tobi-wan-kenobi:
Looks good to me, thanks! Also grateful for the bash improvement, not my strongest suit ;)

One idea for future improvement: It *might* make sense to ensure that the binary is executable, in the future. Also, also, for installations that don't have iwgetid, it might make sense to test for some basic binary that is present all the time (`ls` or something like that). I suspect that in Travis, it'll just always fall into the "None" case.

donfranio:
Hi @tobi-wan-kenobi,
sorry for checking back that late.

1. You are correct, the test would be more precise, if it also checked the x-bit in the file-system.
2. If you would test the existence of ```ls``` then you would only test the test, right? So I would rather go into the direction to run all tests also on a debian-system with wireless-tools installed (of course in some container). This way you find such error in the pipeline already if something changes in debian itself.
3. (a fresh thought) In my opinion it could be useful, if the user recognized, that bumblebee-status is not able to show the essid at all.

What do you say?

donfranio:
Hi again,

So it sure is no problem to iterate in the test over a list of (maybe) used binaries and adding something that needs to be there is surely easy.
Maybe adding some garbage that should never find a binary is a good idea as well. So if you had a second testing environment in the future the "none" path would stay checked.
I will put the stuff into a PR later.
Cu

PR #289, opened 2018-09-04
extra tests for util.which

donfranio:
splitted the linter-stuff in an extra commit for the case that you don't like it ;-)


tobi-wan-kenobi:
Looks perfectly fine to me, many thanks! :+1:

PR #291, opened 2018-09-08
 Add module arch-update 

lucassouto:
Module that show packages to update

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Modules and commandline utilities are only required for modules, the core itself
 * power (for the module 'battery')
 * dbus (for the module 'spotify')
 * i3ipc (for the module 'title')
+* pacman-contrib (for module 'arch-update')
 
 # Required commandline utilities
 
--- a/bumblebee/modules/arch-update.py
+++ b/bumblebee/modules/arch-update.py
@@ -6,6 +6,7 @@ import bumblebee.input
 import bumblebee.output
 import bumblebee.engine
 
+
 class Module(bumblebee.engine.Module):
     def __init__(self, engine, config):
         widget = bumblebee.output.Widget(full_text=self.utilization)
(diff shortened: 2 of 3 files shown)


tobi-wan-kenobi:
Hello and many  thanks for the PR!

According to Travis, we will need to work a bit on the module to make it compatible with older Python versions (see https://travis-ci.org/tobi-wan-kenobi/bumblebee-status/builds/425984966?utm_source=github_status&utm_medium=notification), but that shouldn't be a big deal.

Two more general question, however: 
1. At least on my Arch installation, `checkupdates` is not available, is that something that has to be installed separately?
2. On a first glance, this seems to provide a functionality similar to the module `pacman`, is that right or am I overlooking something?

Thanks a lot in advance!

lucassouto:
Hi @tobi-wan-kenobi, I will make the change to be able to run in other versions of python.

Answering your questions:
1 - checkupdates is a pacman script, I thought it was built-in, but apparently you have to install the pacman-contrib package.
[pacman-contrib](https://www.archlinux.org/packages/community/x86_64/pacman-contrib/)
[Why use pacman-contrib](https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported)
2 - Yes it is similar to pacman, it was inspired by the Gnome plugin to show notifications of packages to be updated.
[GNOME shell arch-update](https://github.com/RaphaelRochet/arch-update)

tobi-wan-kenobi:
Thanks for the quick reply! Looks very good for me, will merge it right away. There's a big overlap between `arch-update` and `pacman`, but that's not a bad thing, I think it is good if users can choose.

Also, in the long-term, I think I'd like to merge `arch-update`, `pacman` and `dnf` into a single module that shows pending OS updates, but that is *very* future work :)

Many thanks for the contribution!

PR #293, opened 2018-09-11
Add xkcd module

whzup:
Added a module that opens a random xkcd comic in the default browser when clicked.

It's a bit silly but I think it might be fun 😄 Close the PR if you don't want to include it 😆 

It requires `xdg-utils` to work.


tobi-wan-kenobi:
That's a nice one! :+1: 

*Of course* I'll merge the PR right away, I'll probably use the module myself :P

PR #297, opened 2018-09-19
Readd iceberg-rainbow theme

whzup:
As I've been demanded to readd the old iceberg theme [here](https://github.com/cocopon/iceberg.vim/issues/39). I'd like to officially readd it under the new name iceberg-rainbow and with some tweaks on the coloring.


tobi-wan-kenobi:
Niceeee :-)

PR #299, opened 2018-10-05
[module/spaceapi] Add Space API module

rad4day:
Hi, I wrote a small module some time ago and decided it was time to check if it would make sense to add it to the main repository. At the moment it waits 5 minutes before making repeated calls to the api endpoint. I'm not sure if the way how I implemented this timeout is the right way. If you have suggestions,  questions or requests, just let me know <3

---
SpaceAPI is a API for hackerspaces. Hackerspaces announce their state
and location using a predefined json framework. Users can use this data
to make sure the space is opened before they decide to go there.

The module was created because the author felt the need to know if his
hometown's space was open or not.

### DIFF
--- a/bumblebee/modules/spaceapi.py
+++ b/bumblebee/modules/spaceapi.py
@@ -1,0 +1,75 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the state of a spaceapi endpoint
+
+Requires the following libraries:
+    * urllib
+    * json
+    * time
+
+Parameters:
+    * spaceapi.url: String representation of the api endpoint
+    * spaceapi.name: String overwriting the space name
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Very nice, thanks!

Regarding the update interval: If you have a look at e.g. `currency.py` or `github.py`, there's a method called `interval()` in the `Module` base class that you can use during initialization to set the update interval. Hope that still works :P 

I'll merge the PR as it is, since it works just fine.

Many thanks for the contribution, much appreciated!

tobi-wan-kenobi:
:) TBH, I'm not too thrilled with the mixture of decorators (scrolling) and inherited methods, but it was the best I was able to come up with... Glad you like it!

PR #300, opened 2018-10-05
[modules/spaceapi] refactor, add comments, use interval function

rad4day:
As announced in #299, here is the PR using the interval function. 
Also I decided to rename some variables and increase the amount of comments.

This change could be considered breaking, as it changes the unit of time from seconds to minutes.


tobi-wan-kenobi:
Looks good, thanks a lot! I also like the refactoring changes, the naming is definitely improved.

PR #307, opened 2018-10-25
[module/spaceapi] fix freeze on unreachable API

rad4day:
The statusbar was freezing for several minutes if it could not reach
the API endpoint. This is because of a missing timeout statement in
the call to python.requests's get function.

- Added spaceapi.timeout parameter
- Added timeout to requests


tobi-wan-kenobi:
Looks good, thanks! If you want to go for something more fancy, you can have a look at the `pacman` module, which updates its info asynchronously. Having said that, I like your approach a lot for its simplicity.

Thanks for the contribution!

PR #309, opened 2018-10-28
Return 0 as an int rather than a string.

kelp:
This was causing an ocassional crash in bumblebee/engine.py threshold_state
when 'checkupdates' fails, perhaps due to wifi not being up yet.

For me this showed up regularly on login.


tobi-wan-kenobi:
Many thanks for the PR! Looks good -> merged :)

PR #310, opened 2018-10-28
Fix the arch icon.

kelp:
On my current Arch Linux system, arch-update was using this arch icon:


Which I just can't find anywhere. It doesn't work on my system with any
of the full patched nerd-fonts installed.

Switching to this icon works:


I'm not sure what the original intention was with that icon/glyph?


kelp:
Realized the intention here was to have a cloud icon. f381 is cloud download alt in FontAwesome. https://origin.fontawesome.com/icons/cloud-download-alt?style=solid

PR #311, opened 2018-10-29
Remove immediate param in github and hipchat update function

fredj:
Not needed after 3638aa24203138476a2a5c8a02bffe9bdc7f27f5


tobi-wan-kenobi:
Good catch, thanks!

PR #313, opened 2018-10-31
[module/weather] Use json function from requests response

fredj:
Small code simplification: use the json function instead of using the json library


tobi-wan-kenobi:
As always, a very nice change - thanks a lot!

PR #314, opened 2018-11-01
Fixed status for unknown charge state, added option to weather to hide city

alexmohr:
Battery showed charging state when it was unknown. This sets the correct state to the block.
Added option to hide city in weather with  weather.showcity="false"

### DIFF
--- a/bumblebee/modules/datetime.py
+++ b/bumblebee/modules/datetime.py
@@ -30,7 +30,6 @@ class Module(bumblebee.engine.Module):
     def __init__(self, engine, config):
         super(Module, self).__init__(engine, config,
             bumblebee.output.Widget(full_text=self.get_time))
-        self._app = self.parameter("open", "calendar")
         engine.input.register_callback(self, button=bumblebee.input.LEFT_MOUSE,
                                     cmd=self._app)
         self._fmt = self.parameter("format", default_format(self.name))
--- a/bumblebee/modules/weather.py
+++ b/bumblebee/modules/weather.py
@@ -33,6 +33,7 @@ class Module(bumblebee.engine.Module):
         self._location = self.parameter("location", "auto")
         self._city = self.parameter("location", "")
         self._unit = self.parameter("unit", "metric")
+        self._showcity = self.parameter("showcity", "true")
         self._valid = False
         self.interval(15)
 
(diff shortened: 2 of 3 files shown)


bumblebee/modules/weather.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, engine, config):
         self._location = self.parameter("location", "auto")
         self._city = self.parameter("location", "")
         self._unit = self.parameter("unit", "metric")
+        self._showcity = self.parameter("showcity", "true")
```
fredj: You could use the `bumblebee.util.asbool` function here:
```python
self._showcity = bumblebee.util.asbool(self.parameter("showcity", True))
```

bumblebee/modules/datetime.py
```diff
@@ -30,6 +30,9 @@ class Module(bumblebee.engine.Module):
     def __init__(self, engine, config):
         super(Module, self).__init__(engine, config,
             bumblebee.output.Widget(full_text=self.get_time))
+        self._app = self.parameter("open", "calendar")
```
tobi-wan-kenobi: All modules automatically have a parameter called "left-click" for binding against executables. My suggestion is to keep "calendar" as the default (by registering the callback like below), but to not make the application itself configurable (i.e. please remove the parameter "open"). Otherwise, there'd be two ways to achieve the same thing (datetime.open and datetime.left-click).

alexmohr: Got it, I'll change that :)

alexmohr: Actually I do not quite understand why you don't want to have the app configurable. I'd like to use google calendar as click action and it's rather ugly to edit the python files instead using a configuration value.

tobi-wan-kenobi: Yes, "left-click" *should* be available for each module. It might be a bit tricky with the "datetime" module, as it has aliases. So, the following should work:

`bumblebee-status -m date -p date.left-click=calendar`
`bumblebee-status -m datetime -p datetime.left-click=calendar`
`bumblebee-status -m time -p time.left-click=calendar`

If it doesn't please let me know, 'cause that is a bug :)

tobi-wan-kenobi:
Thanks a lot for your contribution, PR looks good!

PR #315, opened 2018-11-01
Added Frequency module

alexmohr:
This adds a basic module to watch cpu frequency


alexmohr:
Codeclimate is false positive in my opinion. 
I'll fix the broken tests in the next couple days. I'll reopen the pull request after that.

tobi-wan-kenobi:
Sounds good to me! CodeClimate, I don't care too much about, it's mostly to keep an eye on overall code quality, but if you can make the Travis tests green, that would be much appreciated.

One more thought: The `sensors` module also shows the current frequency, might that fit your requirements?

PR #316, opened 2018-11-01
Added support for date time click

alexmohr:
Date is now clickable :)


tobi-wan-kenobi:
Hi and thanks for your PR!

I see the same changes in #314, was that by accident? I have commented them there, I think adding "calendar" as default app for "datetime" makes sense. There is, however, no need for the parameter "open" to select the application, because *every* module has a generic parameter called "left-click" for binding against applications (e.g. "-p date.left-click=calendar").

alexmohr:
Going to close this pr due to the same changes in #314 
I messed up my branches and wanted to seperate these changes.

PR #317, opened 2018-11-03
Improved battery module for 2 batteries

alexmohr:
I am not quite sure if this is something you want. But the module was not very good when using 2 batteries for example in thinkpads.
The screenshots show different configurations.
Feedback is welcome.

2018-11-03-175455_162x23_escrotum
2018-11-03-175435_215x26_escrotum
2018-11-03-175414_196x21_escrotum
2018-11-03-175345_257x25_escrotum

### DIFF
--- a/bumblebee/modules/battery.py
+++ b/bumblebee/modules/battery.py
@@ -85,7 +85,7 @@ class Module(bumblebee.engine.Module):
                 and self.getCharge(widget) == "Discharging":
             output = "{} {}".format(output, self.remaining())
 
-        if bumblebee.util.asbool(self.parameter("showdevice", True)):
+        if bumblebee.util.asbool(self.parameter("showdevice", False)):
             output = "{} ({})".format(output, os.path.basename(widget.name))
 
         return output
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hm, that's not something I've ever tried, as I don't have a machine with 2 batteries :-( But the code looks like this *should* work. The only restriction I can see (without being able to test) seems to be that all battery widgets have to look identical, right?

Alternatively, maybe you could use aliases to make that work? Like this:

`bumblebee-status -m battery:bat0 battery:bat1 -p bat0.device=BAT0 -p bat1.device=BAT1`

Just out of curiosity, why did you close the PR?

Thanks for your input!

alexmohr:
I closed the PR because I broke the tests and wanted to fix them first. I wouldn't have thought you are that fast. 

If your requirement is that all widgets should look identical this PR is out of scope anyway. When using 2 batteries you would have the runtime in both widgets. With this change it is only in the battery which is discharging, as you can see in the screenshot. 

With your suggestion I'd still have the doubled remaining runtime.

alexmohr:
2018-11-03-183543_245x18_escrotum

This is a screenshot of the original module

tobi-wan-kenobi:
Ah, got it. Yes, that makes a lot of sense. The changes itself look OK to me, so I'll just go ahead and merge them.

Many thanks!

PR #325, opened 2018-11-16
Set state to critical if volume is set over 100%

dustypaws:
..."warning" would suffice but eh!

Useful for headphone users. Or at least I think so... :)


tobi-wan-kenobi:
Nice change, thanks! I like the use of critical, as otherwise, it would clash with muted.

Also, I like that mute continues to take precedence over the >100%.

PR #329, opened 2018-11-17
Fix the remaining time so that it displays the time for all batteries and not just one battery

martindoublem:
Added the lines so that the battery calculations can be done for all batteries. Done for the same way as for batteries total capacity calculation.


tobi-wan-kenobi:
Looks good, thanks a lot!

PR #332, opened 2018-11-24
1.6

RiadhAmami:
xbacklight has a lot of issues with a lot of hardware lately so i got a better alternative light-git and it work fine i changed the brightness.py and it work fine (just changed the command from xbacklight to light)
[br.tar.gz](https://github.com/tobi-wan-kenobi/bumblebee-status/files/2612085/br.tar.gz)

### DIFF
--- a/bumblebee/modules/memory.py
+++ b/bumblebee/modules/memory.py
@@ -50,7 +50,10 @@ class Module(bumblebee.engine.Module):
                 if tmp[2] == "mB": value = value*1024*1024
                 if tmp[2] == "gB": value = value*1024*1024*1024
                 data[tmp[0]] = value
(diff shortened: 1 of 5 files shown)


tobi-wan-kenobi:
Hello and thanks for the PR! The change in the attached tar.gz looks OK (though it needs a bit extension), just a few remarks:

1. I think something went wrong with the PR:
   - It merges the `1.6` release branch instead of `master`
   - It contains various changes from master, but *no* changes to `brightness.py`
2. What would be nice for `brightness.py` would be a way to use `light` if it is available and fall back to `xbacklight` otherwise, in order to maintain backwards compatibility.

What do you think?

tobi-wan-kenobi:
Please let me know whether this change satisfies your request! Thanks for the suggestion!

PR #335, opened 2018-11-26
added new battery type

927589452:


tobi-wan-kenobi:
Thanks for the PR!

Looks good to me (though I don't have /sys/class/power_supply/battery on my system....)

PR #337, opened 2018-12-04
[modules/layout-xkb] use XKeyboard to set layout

MartinWallgren:
XKeyboard has support for setting the layout directly. There is no need
to explicitly execute an external command.


tobi-wan-kenobi:
Awesome, looks good to me. Many thanks for the PR!

MartinWallgren:
Glad to help. I was actually trying to mimic you when I wrote this https://github.com/MartinWallgren/next-layout when I noticed that there was an easier way.

PR #339, opened 2018-12-17
Added "scrolling" config options to spotify and deadbeef modules

WORD559:
This allows the text in the module to scroll across and loop round at a set speed. By default the behaviour is disabled, but can be enabled with the spotify.scrolling or deadbeef.scrolling options. Particularly useful for small screens and long song names.

### DIFF
--- a/bumblebee/modules/deadbeef.py
+++ b/bumblebee/modules/deadbeef.py
@@ -14,9 +14,10 @@ Parameters:
                          Available values are: {artist}, {title}, {album}, {length},
                                                {trackno}, {year}, {comment},
                                                {copyright}, {time}
-    * deadbeef.previous: Change binding for previous song (default is left click)
-    * deadbeef.next:     Change binding for next song (default is right click)
-    * deadbeef.pause:    Change binding for toggling pause (default is middle click)
+    * deadbeef.previous:  Change binding for previous song (default is left click)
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
That's a good idea! Looks good to me, will merge it like this, thanks!

Just one question beforehand: Any specific reason for not using @scrollable (see cmus.py, for example)? Except for the fact that the default behavior would be changed.

WORD559:
Just ignorance, I'm afraid. I wasn't aware it already existed, so I just wrote my own implementation.

WORD559:
Yeah no worries, I'll take a look next chance I get.

tobi-wan-kenobi:
Great, many thanks! I can also add it to the code myself, in a branch, if that helps.

⁣Sent from TypeApp ​

WORD559:
Okay, I've managed to get them to work using @scrollable, and I've instead modified that to get the behaviour I was after.
Since @scrollable seemed to make the text "bounce" back and forth, I've added a <module>.scrolling.bounce parameter to the decorator. This takes either 1 or 0 (1 by default, to preserve existing behaviour) and changes whether or not the text bounces back and forth, or whether it just gets to the end and jumps back to the start.
Additionally, I've added <module>.scrolling.speed, which allows the user to change the scrolling speed. 

Hopefully this is helpful! I'll add the changes in a different PR.

PR #340, opened 2018-12-24
[modules/spaceapi] Fix ArgumentException

rad4day:
I'm a little embarrassed to admit that this actually happened and went unnoticed for so long :/

---
This ArgumentException was caused by me failing to rename one occurence
of a parameter when refactoring code. This went under my radar as the
API I'm usually testing against, was offline at that time, too. So I
expceted to see an Error. Just not this one.

Well, the fix comes late, but better late than never :)

Merry Christmas :christmas_tree: 

### DIFF
--- a/bumblebee/modules/spaceapi.py
+++ b/bumblebee/modules/spaceapi.py
@@ -67,7 +67,7 @@ class Module(bumblebee.engine.Module):
 
     def update(self, widgets):
         try:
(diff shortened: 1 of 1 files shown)


rad4day:
Nevermind, the code is not at a level I'd like it to be right now anyways. Will rewrite the module in the next days.

tobi-wan-kenobi:
Thanks, looking forward to it & merry christmas to you, too!

PR #343, opened 2018-12-28
Modified behaviour of bumblebee.output.scrollable to allow for custom…

WORD559:
… scroll speed and toggling "bouncing" of text, and applied it to deadbeef and spotify

### DIFF
--- a/bumblebee/modules/deadbeef.py
+++ b/bumblebee/modules/deadbeef.py
@@ -2,13 +2,10 @@
 
 """Displays the current song being played in DeaDBeeF and
 provides some media control bindings.
-
 Left click toggles pause, scroll up skips the current song,
 scroll down returns to the previous song.
-
 Requires the following library:
     * subprocess
-
 Parameters:
     * deadbeef.format:   Format string (defaults to "{artist} - {title}")
                          Available values are: {artist}, {title}, {album}, {length},
@@ -17,7 +14,6 @@ Parameters:
     * deadbeef.previous: Change binding for previous song (default is left click)
     * deadbeef.next:     Change binding for next song (default is right click)
     * deadbeef.pause:    Change binding for toggling pause (default is middle click)
-
     Available options for deadbeef.previous, deadbeef.next and deadbeef.pause are:
         LEFT_CLICK, RIGHT_CLICK, MIDDLE_CLICK, SCROLL_UP, SCROLL_DOWN
 """
@@ -25,6 +21,8 @@ Parameters:
 import bumblebee.input
 import bumblebee.output
 import bumblebee.engine
+
+from bumblebee.output import scrollable
 
 try:
     import subprocess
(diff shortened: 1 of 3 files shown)


WORD559:
Ah damn, the spotify module doesn't seem to like using @scrollable. I'll check it out once I've eaten, but for now this isn't safe to merge.

WORD559:
Yeah, the festivities are the reason it's taken me so long to take a look at @scrollable. I'll take a look at it ASAP and see if I can get it working.

If you want to modify the PR to remove the changes to the spotify module, that's fine -- from what I've seen, it works perfectly with deadbeef.

tobi-wan-kenobi:
Hm, diff looks fine to me. Not sure I can change somebody else's PR, tbh (but then, I'm not exactly an expert on github...)

Probably it's something in the spotify module directly, and would also have broken with the original scrollable code :)

I'd be fine with merging the PR if you just disable scrollable for spotify at the moment.

Cheers!

WORD559:
Okay, I'll do that. I've logged the error message that occurs, so I'll post it below for future reference and we can try and figure the problem out at some point.

```
  File "/home/owner/.config/i3/bumblebee-status/bumblebee-status", line 80, in <module>
    main()
  File "/home/owner/.config/i3/bumblebee-status/bumblebee-status", line 74, in main
    output.draw(widget, error)
  File "/home/owner/.config/i3/bumblebee-status/bumblebee/output.py", line 125, in draw
    padding = self._theme.padding(widget)
  File "/home/owner/.config/i3/bumblebee-status/bumblebee/theme.py", line 110, in padding
    return self._get(widget, "padding", "")
  File "/home/owner/.config/i3/bumblebee-status/bumblebee/theme.py", line 238, in _get
    state_themes.append(self._get(widget, state, {}))
  File "/home/owner/.config/i3/bumblebee-status/bumblebee/theme.py", line 256, in _get
    if mod and not mod.parameter("is-unittest"):
  File "/home/owner/.config/i3/bumblebee-status/bumblebee/engine.py", line 114, in parameter
    value = self._config["config"].get(name, default)
TypeError: 'Config' object has no attribute '__getitem__'
```

WORD559:
I believe, for future reference, if you wish to modify my PR, you should have write permissions for the branch the PR is based on (in this case, my `scrolling` branch). You should be able to add commits to that branch to modify the PR, so long as I have "Allow edits from maintainers." ticked (which I usually do).

tobi-wan-kenobi:
Awesome, thanks!

As to the error: That *seems* like a bug in the engine that slipped in at some point in time. Let me run a few tests and I'll hopefully push a fix soonish.

tobi-wan-kenobi:
OK, I've dug into the code a bit now, and have to admit: No clue what is going on. I cannot reproduce the issue (I have to "fake" it in the spotify module, as I don't have a spotify account, but still...)

Also, I don't see how the spotify module could work differently in this case than, say, deadbeef.

Can you maybe get a debug output that shows the contents of `self._config`? It *should* be a dict containing a "name" and a "config" field, but I suspect it's just a Config() object. But I don't get how that is possible, as the only initialization happens in `engine.py:248`.

Any kind of debug output would be *most* helpful :)

PR #348, opened 2019-01-13
Update dependencies in layout-xkb

mkwardakov:


tobi-wan-kenobi:
Good catch, thanks for the PR!

PR #349, opened 2019-01-14
Make mocp understand formats

mkwardakov:
Slight init optimization


tobi-wan-kenobi:
Looks good to me, thanks for the contribution! I particularly like changesets that remove more lines than they add :)

PR #358, opened 2019-01-31
Remove hipchat module

fredj:
Hipchat will be discontinued after February 15th, 2019.


tobi-wan-kenobi:
Oh, didn't know about that, what a pity. Thanks for the PR!

PR #361, opened 2019-02-14
Added module for prime video cards (nvidia/intel hybrid computers).

jeffeb3:
My new computer has a prime setup, running either nvidia or intel graphics options. This lets me see which one is running, and change it with a click.

prime-select query prints either intel or nvidia.
sudo prime-select intel selects the intel card (good for longer battery)
sudo prime-select nvidia selects the nvidia card (good for performance)

It's missing the fancy theme icons. But the user can put whatever in the settings strings. I'm not sure how those work, if they glob replace words like cpu or what.

I verified this locally. Tested the --list modules output, left, right clicks, and the two parameters.

### DIFF
--- a/bumblebee/modules/prime.py
+++ b/bumblebee/modules/prime.py
@@ -1,0 +1,69 @@
+# pylint: disable=C0111,R0903
+
+"""Displays and changes the current selected prime video card
+
+Left click will call 'sudo prime-select nvidia'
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good, thanks! There's one minor thing, the return value of the full_text method should be a string, but I'll fix that after merging.

Many thanks for the PR!

PR #374, opened 2019-03-30
Clarify testjson.sh output, fix ascii.json and ionicons.json

tri2820:


tobi-wan-kenobi:
Thanks for the contribution, the diff looks good!

PR #375, opened 2019-04-06
[modules/hostname] add hostname module

varkokonyi:
Hi.
I just decided to add this simple module, just for fun.
I'm not sure it's worth adding, but here it is


tobi-wan-kenobi:
Looks good, many thanks for the PR!

PR #376, opened 2019-04-13
Detect wireguard interfaces as tunnel type

TheEdgeOfRage:
I started playing around with wireguard a bit and it annoyed me that the interface was marked as wireless. So I added just a simple check for the default interface name which is wg<number>


tobi-wan-kenobi:
Hi @TheEdgeOfRage , welcome back :)

Nice PR, thanks!

PR #379, opened 2019-04-22
Change to calculate estimate time and percentage

stoeps13:
I had issues with the provided way the script analysis my batteries in my lenovo T480 on Fedora 29 and 30. The remaining capacity and time was completely wrong, so I changed the calculation to float.
Now it shows the right percentage and estimated time left. Not tested with other distributions.

### DIFF
--- a/bumblebee/modules/battery_all.py
+++ b/bumblebee/modules/battery_all.py
@@ -27,7 +27,6 @@ class Module(bumblebee.engine.Module):
         # TODO: list all batteries
         self._batteries.append("/sys/class/power_supply/BAT0")
         self._batteries.append("/sys/class/power_supply/BAT1")
-        self._batteries.append("/sys/class/power_supply/battery")
 
         super(Module, self).__init__(engine, config, bumblebee.output.Widget(full_text=self.capacity))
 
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hello,

thanks a lot for the PR! I agree that converting to float makes sense, doing it the way I did until now is a bug. I'll fix that right away.

However, in your PR, you also completely comment out the fallback remaining time calculation, and that would break backwards compatibility for some people.

If the display is wrong for you and you don't want to see the calculated remaining time, can I suggest that we add a toggle that simply hides this information?

Thanks!

tobi-wan-kenobi:
I've now committed your fix in e0a7ca5. Also, I've noticed that the flag I mentioned is already there, I believe (`battery_all.showremaining`), if you set that to `false`, I think that does what you want.

stoeps13:
Thanks, but the fallback does not work! I want to see remaining, but when I check the if ... Else part and compare with battery.py, there is something weird. 

When I'm on battery the else part always jumps in. First if == -1.0, second == -2.0, so always else is true. The calculation in else is not right. Power module calculate 10-12 hours which is right, this fallback shows around 40 min.

Will check again what happens there.

tobi-wan-kenobi:
Ah, I get your point now. Yes, you are correct, this is currently not working as intended.

tobi-wan-kenobi:
Looking at the code a bit, I see you are completely right: This path is wrong, and there's no simple way to cleanly do a fallback calculation. I've followed your advice and removed the code in question.

If the module is fine as it is for you, I'd close the PR now.

Thanks!

stoeps13:
Sure, looks great. I'm fine with closing this. Sorry for the delay, but was on vacation.
Greetz

PR #385, opened 2019-05-12
Added pi-hole module

bbernhard:
This pull request adds a pi-hole module to bumblebee. 

The pi-hole module shows the number of blocked ads (on a per-day basis) and allows to enable/disable pi-hole per click. 

I am by no means a user experience expert...so any UI improvement suggestions are really welcome ;) 

btw: pi-hole has a pretty nice option that allows you to disable it for x minutes and then automatically enable it again. This is all done via URL query parameter (e.q: ` ?disable=10`  to disable it for 10 seconds), so there is no need to remember the elapsed time in the module. 

I thought about implementing that too, but still need to check how/if it's possible to dynamically set the duration in bumblebee. An alternative would probably be to set the duration in the config..but I think that might be too inflexible for some use cases.

### DIFF
--- a/bumblebee/modules/pihole.py
+++ b/bumblebee/modules/pihole.py
@@ -1,0 +1,68 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the pi-hole status (up/down) together with the number of ads that were blocked today
+Parameters:
+    * pihole.address     : pi-hole address (e.q: http://192.168.1.3)
+    * pihole.pwhash      : pi-hole webinterface password hash (can be obtained from the /etc/pihole/SetupVars.conf file)
+"""
+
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thanks for the PR! As I have a pihole at home myself, I'll most definitely merge this right away :)

A couple of suggestions:
* Right now, you update the module every cycle, which generates an HTTP request every time (1 per second, by default) - if you use the `ìnterval` methods from the framework (see `weather.py` for an example), you can significantly reduce that. Note that when you click the module, it will still get updated
* For adding controls, you can look at `cmus.py`, which creates separate widgets for play/pause, forward, rewind, etc.

Many thanks again for the contribution!

PR #388, opened 2019-05-20
Add support for city ids to weather module.

david-perez:
Some cities (e.g. in different countries) have the same names (e.g. [Oxford](https://openweathermap.org/find?q=Oxford)). Providing a city name to the weather module would previously display the first returned result by openweathermap's API.

This commit allows city ids to be provided to the weather module. If a location passed to the weather module only contains numbers, it will be interpreted as a city id. City ids are those handled by openweathermap's API to uniquely identify cities.

See https://openweathermap.org/current#cityid for details.

### DIFF
--- a/bumblebee/modules/weather.py
+++ b/bumblebee/modules/weather.py
@@ -7,8 +7,9 @@ Requires the following python packages:
     * requests
 
 Parameters:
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks for the PR! Oooohhh, that's a useful change and implemented in a very nice way!

PR #389, opened 2019-05-20
convert title.max parameter to int

Stoope:


tobi-wan-kenobi:
Thanks for the contribution, good catch!

PR #392, opened 2019-06-02
fixed bug in bumblebee popup implementation

bbernhard:
Please correct me if I am wrong, but I _think_ there is a typo in the popup implementation. It should be `command=callback` instead of `command=click_callback`.


tobi-wan-kenobi:
Yeah, that looks like a bug to me, too. Thanks for the PR!

PR #394, opened 2019-06-04
Added datetime module with timezone support

frankzhao:
Adds a new module `datetimetz` that extends the `datetime` module to display the current time in various timezones.

A comma separated list of IANA timezones can be specified in the parameters for multiple timezone support. Click the widget to cycle through the list.

image

Sample config for above:
```
bumblebee-status -m datetimetz -p datetimetz.timezone="Australia/Sydney,Pacific/Auckland,America/Los_Angeles"
```

Note: I created this as a separate module since it requires additional dependencies (`pytz`, `tzlocal`) and I did not think the default (simpler) datetime module should require optional dependencies.


tobi-wan-kenobi:
Very cool module, thanks a lot for the contribution!

PR #400, opened 2019-06-23
changed epoll to poll, updated README for FreeBSD

Crupsp:
Commit message says it all, I got bumblebee working under FreeBSD and wanted to spare others the hassle, hope this PR is welcome.
Thanks for bumblebee, I love it with many others!

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ Explicitly unsupported Python versions: 3.2 (missing unicode literals)
 # Arch Linux
 $ sudo pacman -S awesome-terminal-fonts
 
+# FreeBSD
+$ sudo pkg install font-awesome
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thanks a lot for the PR!

Nice to know that bumblebee works in FreeBSD, too. Reg. poll: Makes a lot of sense, I just need to investigate why the tests fail. I suspect it's due to the change from edge-triggered to whatever poll does.

I'll look into it and merge the PR as soon as I have a solution. It might take a few days, but I promise I'll get around to it.

tobi-wan-kenobi:
OK, looks good, seems to be mostly an error around the mocking of poll vs. epoll, I'll merge now.

PR #404, opened 2019-06-30
[modules/arch-update] add format parameter

kockahonza:
Just added `arch-update.format` parameter so that the text can be easily changed and fixed one indent (I'm assuming it wasn't intentional).


tobi-wan-kenobi:
Many thanks for the contribution, looks totally fine!

PR #408, opened 2019-07-10
[module] Basic alternate network traffic module

izn:
In this PR I'm proposing an alternate network traffic module.

Main differences between this and the default traffic module:
- No extra configurations
- Easy to use
- Get information from the default network adapter

#### Usage

No extra configuration needed. Just `-m network_traffic`.

I have been using this module for 3 months+ and it's working very well.

#### Dependencies

* psutil
* netifaces

#### Tests

I would like to implement some unit tests, but it's kinda hard to mock a network adapter. I think I will need to rewrite some classes using dependency injection later. I hope I could cover this module in future.

#### Screenshot

ss

### DIFF
--- a/bumblebee/modules/network_traffic.py
+++ b/bumblebee/modules/network_traffic.py
@@ -1,0 +1,42 @@
+"""Displays network traffic
+"""
+
+import psutil
+import netifaces
+import bytefmt
+
+import bumblebee.input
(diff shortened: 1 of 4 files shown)


tobi-wan-kenobi:
Very cool contribution, many thanks!

PR #411, opened 2019-07-23
Add http status module

valkheim:
This pull request adds an http status checker module to bumblebee.

The http status module allows to check an http status on a regular basis from a head request.
Parameters:
- **http_status.target**: the target to retrieve the http status from.
  *e.g.* http://example.org
- **http_status.expect**: you can expect a particular http status ;
  *e.g.* let's say I expect my front page to return a 200.
- **http_status.label**: adds a prefixed label on the widget.
  *e.g.* `example.org: 200`

Some screenshots
- successful status:

- expected 400 for example.org but got 200:
 

- unknow status (due to network issue or target mistyping):

There is some tests included.


tobi-wan-kenobi:

PR #412, opened 2019-07-28
system module

bbernhard:
Simple module that allows to shutdown/reboot your system. 

out of interested: Is it somehow possible to access the the module's iconset from the currently selected theme? 

I would like to add some icons to the bumblebee menu entries. so, basically like this:
Selection_017


tobi-wan-kenobi:
Change looks good - many thanks for the PR!

Regarding the iconset: Not (yet), will look into that during the weekend!

PR #413, opened 2019-07-30
Update README.md

927589452:
Add line breaks in statuscommand example
(this is officially supported in i3:
https://i3wm.org/docs/userguide.html 4.31. )
this adds a great deal of readability to the config


tobi-wan-kenobi:
Thanks for the PR! Yes, that's a good suggestion that improves readability!

PR #414, opened 2019-07-31
TWMN Module

Pseudonick47:
Added module for switching twmn on/off. Also, formatted a few json files


tobi-wan-kenobi:
Great, many thanks for the PR!

PR #419, opened 2019-08-12
Sun module

lonesomebyte537:
Module that shows the sunrise and sunset time for your location.


tobi-wan-kenobi:
Many thanks for the PR, looks good!

PR #420, opened 2019-08-15
Initial version of RSS news ticker

lonesomebyte537:


lonesomebyte537:
Tobi, 2 more codeclimate issues left. Can you check and confirm these are acceptable. Don't see a nice way solving these. You?

tobi-wan-kenobi:
Changes look good, many thanks for a very cool module!

PR #421, opened 2019-08-16
Bugfix in parameter parser

lonesomebyte537:
This bug prevented equal signs from being used in parameter values, e.g.
-p url="http://www.foo.org/?a=5"


tobi-wan-kenobi:
Great find, thanks!

PR #422, opened 2019-08-19
Add log out, switch user, lock, suspend and hibernate options to module system

adario7:
Modifications to the 'system' module to enable more options through systemd: log out, suspend and hibernate


tobi-wan-kenobi:

PR #424, opened 2019-08-21
Add location options to the redshift module

basvdheuvel:
I noticed my computer's memory fill up because redshift uses geoclue2 by default. It turns out redshift also supports manual location configuration, so I added an option to manually set a latitude and a longitude for redshift, or to use ipinfo, a method used by some other location-dependent modules.

### DIFF
--- a/bumblebee/modules/redshift.py
+++ b/bumblebee/modules/redshift.py
@@ -12,11 +12,13 @@ import bumblebee.input
 import bumblebee.output
 import bumblebee.engine
 
+
 def is_terminated():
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Oh, yes, geoclue also killed my box occasionally. I actually configured redshift via its own config file, but I am very much open to the idea of providing that interface in bumblebee directly.

After all, one less configuration file needed :)

Many thanks for the addition!

I'll just make one small addition: I'll make the requests module optional by catching the import error. Hope this is fine for you.

PR #428, opened 2019-08-23
added a line at 106

Damicristi:


tobi-wan-kenobi:
Yes, that makes more sense in terms of formatting, thanks for the conribution!

PR #429, opened 2019-08-24
performance improvements on module 'title'

adario7:
- reformat: renamed private methods and variables to start with an underscore
- performance: parse the arguments one time, at the beginning
- performance: update the title through an event listener, as opposed to polling it every update()

### DIFF
--- a/bumblebee/engine.py
+++ b/bumblebee/engine.py
@@ -63,7 +63,7 @@ class Module(object):
     def widgets(self, widgets=None):
         """Return the widgets to draw for this module"""
         if widgets:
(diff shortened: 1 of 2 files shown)


adario7:
looks like i have issues on one of the few lines i have not touched... xD
any idea on how the 'complexity' can be reduced?

tobi-wan-kenobi:
Looks good for me, thanks for the PR!

Regarding the complexity: No worries, that's more "FYI" and not preventing a PR from being merged.

I've not fully calibrated CodeClimate, so it tends to report problems quite early :)

PR #431, opened 2019-08-28
find aliases without importing modules

adario7:
currently, `engine.py` imports all the modules to load their aliases, which takes much time when bumblebee is starting and spits out errors for missing dependencies when the modules aren't actually used. to prevent this I separated the aliases form the python modules, so they can be loaded apart from them.

### DIFF
--- a/bumblebee/engine.py
+++ b/bumblebee/engine.py
@@ -16,6 +16,16 @@ try:
 except ImportError:
     from configparser import RawConfigParser
 
+def all_modules():
+    """Return a list of available modules"""
+    result = []
+    path = os.path.dirname(bumblebee.modules.__file__)
+    for mod in [name for _, name, _ in pkgutil.iter_modules([path])]:
+        result.append({
+            "name": mod
+        })
+    return result
+  
 class Module(object):
     """Module instance base class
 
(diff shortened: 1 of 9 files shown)


tobi-wan-kenobi:
Thanks a lot for the PR! On first glance, this looks like a great improvement. I'll try to thoroughly review the PR this weekend & merge it then.

tobi-wan-kenobi:
Hi,

so, the diff looks perfectly fine and I think it's a great improvement. However, there's still dangling references to all_modules - one in config.py, and 2 in the tests.

I'm thinking that maybe it's best not to delete the `all_modules()` method, and instead just change the `_read_aliases()`, what do you think? Also, it might be easier to define all aliases in a single JSON file, instead of finding them via filesystem operations.

What's your opinion on that?

adario7:
Sorry, i was not aware of the other uses of `all_modules`, i'll put it back immediately.
About the json, i think it is quite overkill to use it just to load a list of names. Surely it would be more efficient to read only one file as opposed to many, but i did it this way to leave each module completely separated from the others (plus not all modules have aliases, so it not *that many* files to open).

tobi-wan-kenobi:
Looks good to me, many thanks for incorporating those changes!

PR #433, opened 2019-08-30
[battery_all] Fixed the battery path and used os to solve for the corrects paths

martindoublem:
Fixes the point, I raised in my issue


tobi-wan-kenobi:
Hi, thanks for fixing this!

PR #435, opened 2019-09-05
Improve screensaver and dpms handling in caffeine module

### LINKED ISSUE #426: Screen still blanks even if caffeine is enabled (for certain setups)
#### Summary
Affected module: caffeine.py

Description:
The screen still blanks after 10 minutes, even if caffeine is active. The cause of the problem is that DPMS is still activated, which can cause issues on some setups.
A simple solution would be to enable/disable DPMS (+dpms, -dpms) when calling xset in caffeine.py. Though I'm not 100% sure which implications modifying DPMS has on screensaver. Possibly screensaver are intended to handle DPMS on their own, so modifying it directly in caffeine.py might interfere with screensaver.

#### What is the current behaviour and what is the expected behaviour?
Current behaviour: Screen turns off after 10 minutes
Expected behaviour: Screen doesn't turn off when caffeine is activated

#### How to reproduce
I used a default installation of xorg-server under ArchLinux without modifying any DPMS stuff in the X config. I don't use a display manager and run the X server via startx. When enabling caffeine and waiting for 10 minutes without any user interaction, DPMS will blank the screen.

#### Expected behavior
Screen doesn't turn off when caffeine is activated.

cddmp:
Trying to fix #426


tobi-wan-kenobi:
Looks good to me, many thanks. That was a lot of work for sure, I appreciate it!

cddmp:
Thank you :)

PR #436, opened 2019-09-06
add firefox-dark-powerline Theme

donfranio:
powerline-like theme using the colors from the dark theme firefox is shipped with.


tobi-wan-kenobi:
Cool addition, many thanks for the PR!

PR #441, opened 2019-09-23
Add support for nowplaying-tf in deadbeef module

joshbarrass:
This allows more sophisticated queries in the same style as foobar2000's title formatting queries. This massively increases the customisability of the module text as it is generated entirely by deadbeef, rather than by the module, and supports various arithmetic and control operations. 

The module should also properly support unicode text now, regardless of python version.

I've also done a few bits of clean-up on the module where I can, my code isn't always the most readable and emacs loves to complain about it :^)


tobi-wan-kenobi:
Cool, thanks a lot for the PR! Looks good to me :)

PR #443, opened 2019-09-25
Add Docker Support

brianlechthaler:
Added Dockerfile for easier deployment. Does not contain 'CMD pip3 install $packages' directive for module dependency fetching.


tobi-wan-kenobi:
That's a great idea, thanks!

PR #444, opened 2019-09-25
add vault default text option

rotsix:
vault.py uses hard-coded text, now possible to tweak it


tobi-wan-kenobi:
Nice quality-of-life improvement, thanks!

PR #445, opened 2019-09-25
Support parsing zpool output for ZFS >=0.8.0

Ninpo:
Query /sys/module/zfs/version for ZFS version and account for the
additional CKPOINT field in ZFS 0.8.0 and higher.

### DIFF
--- a/bumblebee/modules/zpool.py
+++ b/bumblebee/modules/zpool.py
@@ -65,8 +65,13 @@ class Module(bumblebee.engine.Module):
 
     def _update_widgets(self, widgets):
         # zpool list -H: List all zpools, use script mode (no headers and tabs as separators).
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks a lot for the PR!

That looks like a really valuable change, could you maybe have a look to fix the autotests? A check for whether /sys/module/zfs/version exists and handling a failure (maybe assume a default version?) should do the trick.

Ninpo:
> Thanks a lot for the PR!
> 
> That looks like a really valuable change, could you maybe have a look to fix the autotests? A check for whether /sys/module/zfs/version exists and handling a failure (maybe assume a default version?) should do the trick.

Certainly, I've made the above changes and the Travis builds now pass.

tobi-wan-kenobi:
Perfect, thanks a lot!

Issue #434, opened 2019-09-05
Trying to write a module


karthink:
|    Q        |   A
|------------ | -----
| Version     | Unknown, can't find version number

Couldn't find the version number in the source file.

#### Support Question
I am trying to write a module (pomodoro timer) for bumblebee-status, but I can't get it to work. It shows up on the status bar, but I can't get the timer to start. 

```
# pylint: disable=C0111,R0903

"""Display and run a Pomodoro timer.
Left click to start timer, left click again to pause.
Right click will cancel the timer.
"""

from threading import Timer
import subprocess

import bumblebee.input
import bumblebee.output
import bumblebee.engine

class Module(bumblebee.engine.Module):
    def __init__(self, engine, config):
        widgets = bumblebee.output.Widget(full_text=self.text)
        self.remaining_time = 25
        self.pomodoro = { "state":"OFF", "type": "n/a" }
        self._text = str(self.remaining_time) + "m " + self.pomodoro["type"] + " " + self.pomodoro["state"]
        super(Module, self).__init__(engine, config, widgets)
        engine.input.register_callback(self, button=bumblebee.input.LEFT_MOUSE,
                                       cmd=self.timer_play_pause)
        engine.input.register_callback(self, button=bumblebee.input.RIGHT_MOUSE,
                                       cmd=self.timer_reset)
        
    def text(self,widget):
        return "{}".format(self._text) 
        
    def update(self,widget):
        self._text = str(self.remaining_time) + "min " + self.pomodoro["type"] + " " + self.pomodoro["state"]
    
    def timer_play_pause(self, widget):
        if self.pomodoro["state"] == "OFF":
            self.pomodoro = {"state": "ON", "type": "WORK" }
            self.remaining_time = 25
            self.run_timer()

        elif self.pomodoro["state"] == "ON":
            self.pomodoro["state"] = "PAUSED"
            self.stop_timer()
            
        elif self.pomodoro["state"] == "PAUSED":
            self.pomodoro["state"] = "ON"
            self.run_timer()

    def timer_reset(self, widget):
        if self.pomodoro["state"] == "ON":
            self.pomodoro = { "state":"OFF", "type": "n/a" }
            self.remaining_time = 25
            self.stop_timer()
    
    def run_timer(self, widget):
        self.one_minute_timer  = Timer(60,self.decrement_remaining_time)
        self.one_minute_timer.start()

    def stop_timer(self, widget):
        try:
[...]

PR #447, opened 2019-09-26
[module] New module pomodoro timer

martindoublem:
Created the pomodoro timer module, tested on my machine and worked fine.

Based on the work initialy done by karthink. In issue #434

### DIFF
--- a/bumblebee/modules/pomodoro.py
+++ b/bumblebee/modules/pomodoro.py
@@ -1,0 +1,67 @@
+# pylint: disable=C0111,R0903
+
+"""Display and run a Pomodoro timer.
+Left click to start timer, left click again to pause.
+Right click will cancel the timer.
+"""
+
+import datetime
+
+import bumblebee.input
+import bumblebee.output
+import bumblebee.engine
+
+class Module(bumblebee.engine.Module):
+    def __init__(self, engine, config):
+        widgets = bumblebee.output.Widget(full_text=self.text)
+        self.remaining_time = datetime.timedelta(minutes=25)
+        self.remaining_time_str = "{}min{}s ".format(int((self.remaining_time.seconds / 60)),
+                                                     round((self.remaining_time.seconds/60) % 1*60))
+        self.time = None
+        self.pomodoro = { "state":"OFF", "type": "n/a"}
+        self._text = self.remaining_time_str + self.pomodoro["type"] + " " + self.pomodoro["state"]
+        super(Module, self).__init__(engine, config, widgets)
+        engine.input.register_callback(self, button=bumblebee.input.LEFT_MOUSE,
+                                       cmd=self.timer_play_pause)
+        engine.input.register_callback(self, button=bumblebee.input.RIGHT_MOUSE,
+                                       cmd=self.timer_reset)
+        
+    def text(self, widget):
+        return "{}".format(self._text) 
+        
+    def update(self, widget):
+        if self.pomodoro["state"] == "ON":
+            timediff = (datetime.datetime.now() - self.time)
+            if timediff.seconds >= 0:
+                self.remaining_time -= timediff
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks a lot!

It would be really great to have a pomodoro module in bumblebee. The Travis failure in Python 2.7 looks like a fluke, like it's trying to use bumblebee's datetime module instead of the Python module, but I'll need to investigate that during the weekend in a bit more detail.

martindoublem:
Ideally there is probably a few things we can add and change to make it look good, like the times in config for each, like work and play.

tobi-wan-kenobi:
The module looks fine, I'll just go ahead and merge it :)

karthink:
Thank you Martin!

@tobi: I want to add two things to the module
1. An icon/unicode character. I looked at the spotify and redshift modules to see how they do it, and I can't find the character anywhere.
2. To highlight the display when the timer is running, the way the bluetooth module is lit up when bluetooth is on, or the CPU module is at high usage. But I can't see where to define this behavior.

tobi-wan-kenobi:
Hi @karthink ,

good ideas!

Reg. 1 (icons): Please have a look at the themes in the themes/ subdirectory in general, and in particular at the themes/icons/ subdirectory. There, icons (prefixes and postfixes) for all modules are defined. If you add a definition for "pomodoro", you should get the icons automatically.

Reg. 2 (highlights): That's achieved via the `def state(self, widget)` method of a module. If the module returns "warning", it'll automatically get colored like a warning (typically yellow), if it returns "critical", it will get colored red (or whatever the theme says). You can use arbitrary status strings, but if it's something other than "critical" or "warning", you need to manually adapt the themes to add coloring for that.

Hope that helps!

karthink:
off
work
paused
break

Added color definitions and icons for most themes and all icon packs. It's different colors for each theme to try and blend in, but it's not a perfect match everywhere. The gruvbox and solarized themes will work best.

Also tweaked the text descripton to be shorter, since the icons convey information about the state.

I forked it to make the changes, I want to make a couple more before submitting a pull request.

1. How do I set module parameters from the command line? I checked out nic.py because it has a lot of those, but I'm not seeing how they work.
2. Along the same lines, I want to add a command line parameter that lets the user specify a "bell" command to call when the timer runs out, like `bumblebee-status -m pomodoro -p pomodoro.bell="notify-send Time up"`. How would I do that?

karthink:
Found better icons from the font-awesome pack.
off_000
work_000
paused_001
break_000

martindoublem:
Alright, there are a couple settings (think config) I would like to set:

- Decide wether to display seconds or not.
- Have an option to change Work time and break time

@karthink , did you merge your changes yet?

martindoublem:
Alright, there are a couple settings (think config) I would like to set:

- Decide wether to display seconds or not.
- Have an option to change Work time and break time

@karthink , did you merge your changes yet?

karthink:
@martindoublem No the changes are in my forked repo right now. If you know how to make it configurable you can take over. Looks like we need the following configurable options:

1. A bell command to ring at the end of the work/break periods
2. Work and break durations
3. Time display option for minutes/minutes+seconds.

tobi-wan-kenobi:
@karthink that looks real nice!

Regarding your questions:

1. Parameters can be defined using `self.parameter(<parameter name>, <default value>)`. Check out datetime.py for an example.
2. For notify, there's sample code in caffeine.py. It is really straight-forward.

Hope this helps! Looking forward to the PR a lot!

karthink:
OK I created a pull request.

> 1. A bell command to ring at the end of the work/break periods
> 2. Work and break durations
> 3. Time display option for minutes/minutes+seconds.

All of these are implemented, #3 is fully configurable with a user-supplied string format.

PR #449, opened 2019-09-30
Pomodoro module - improved

karthink:
The Pomodoro module is edited in the following ways:
1. Theme support for all themes + custom awesome-icons. Not sure what to do about ionicons, please have a look at ionicons.json.

2. Fully customizable timer format, any string with"%m" and/or "%s" for minutes and seconds. Omissions are handled gracefully.

3. Customizable work period, break period and notify command for when the timer ends/begins. The notify command uses bumblebee.util.execute and fails when I try something like pomodoro.notify='aplay /path/to/sound/chime.wav'. I don't understand why so I haven't wrapped the call in a `try`..`except` block.


tobi-wan-kenobi:
Looks good to me, many thanks! And since Travis says the tests run fine, I'll just go ahead and merge.

Thanks for this great addition!

PR #450, opened 2019-10-02
parameter to prevent scrollable decorator from fixing the module width

joshbarrass:


tobi-wan-kenobi:
Looks good, even though you *could* use the bumblebee.util.asbool() for making the boolean comparison even nicer :)

Thanks!

joshbarrass:
Would it be worth me submitting another PR to switch it over to `bumblebee.util.asbool()`? I don't want to spam you with unnecessary PRs.

tobi-wan-kenobi:
I'd welcome the PR! Otherwise, I'll just fix that with one of my next commits, both work fine for me.

joshbarrass:
Okay, I'll put a PR together.

PR #452, opened 2019-10-03
Use util.asbool rather than manual true-value parsing

joshbarrass:
https://github.com/tobi-wan-kenobi/bumblebee-status/pull/450#issuecomment-537685091

### DIFF
--- a/bumblebee/output.py
+++ b/bumblebee/output.py
@@ -7,15 +7,14 @@ import json
 import uuid
 
 import bumblebee.store
-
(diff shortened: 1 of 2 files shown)


joshbarrass:
I'm unsure why a seemingly-unrelated change would drop the test coverage for two of the modules. My only assumption is that adding the bumblebee.util import has added more untested code to those files? If I get a chance tonight I'll see if I can write a few more tests for you to bring it back up :)

tobi-wan-kenobi:
Perfect, that looks really good, many thanks!

PR #459, opened 2019-10-20
[module/battery-upower] New battery module UPower

martindoublem:
New battery module getting information from upower instead of the sys kernel folder.


tobi-wan-kenobi:
Awesome, many thanks for the addition!

PR #461, opened 2019-10-27
[modules/shell] Add shell module: Execute command in shell and print …

rrhuffy:
Shell module is for quick and easy monitoring anything achievable from a shell.
For long-running scripts, to avoid bumblebee-status freezes during updates, there is an async parameter for spawning separate thread for updating module output.
Checked with pylint (10/10), tested all examples with Python 3.6.8 and simple 'echo hi' on 2.7.15.
Used '{}'.format(var) syntax instead of 3.6+ f'{var}' to have compatibility with older versions.
No unit tests because I haven't any clue how to write a unit test for it which won't be an integration test (like testing simple echo with pipe to grep for example)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Test Coverage](https://codeclimate.com/github/tobi-wan-kenobi/bumblebee-status/badges/coverage.svg)](https://codeclimate.com/github/tobi-wan-kenobi/bumblebee-status/coverage)
 [![Issue Count](https://codeclimate.com/github/tobi-wan-kenobi/bumblebee-status/badges/issue_count.svg)](https://codeclimate.com/github/tobi-wan-kenobi/bumblebee-status)
 
-**Many, many thanks to all contributors! As of now, 50 of the modules are from various contributors (!), and only 19 from myself.**
+**Many, many thanks to all contributors! As of now, 51 of the modules are from various contributors (!), and only 19 from myself.**
 
 ![Solarized Powerline](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/screenshots/themes/powerline-solarized.png)
 
--- a/bumblebee/modules/shell.py
+++ b/bumblebee/modules/shell.py
@@ -1,0 +1,91 @@
+# pylint: disable=C0111,R0903,W1401
+
+""" Execute command in shell and print result
+
+Few command examples:
+    'ping 1.1.1.1 -c 1 | grep -Po "(?<=time=)\d+(\.\d+)? ms"'
+    'echo "BTC=$(curl -s rate.sx/1BTC | grep -Po \"^\d+\")USD"'
+    'curl -s https://wttr.in/London?format=%l+%t+%h+%w'
+    'pip3 freeze | wc -l'
+    'any_custom_script.sh | grep arguments'
+
+Parameters:
+    * shell.command:  Command to execute
+                      Use single parentheses if evaluating anything inside
+                      For example shell.command='echo $(date +"%H:%M:%S")'
+                      But NOT shell.command="echo $(date +'%H:%M:%S')"
+                      Second one will be evaluated only once at startup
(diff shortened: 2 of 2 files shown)


tobi-wan-kenobi:
Hi,

very cool addition, thanks a lot!

I'll have a look at why the autotests fail, seems to be a mocking error.

One suggestion: It *might* be useful to allow the user to have a larger interval between invocations than the default (such as the `ping` module).

rrhuffy:
By using "shell.interval" user can define the time between updates.
If I understood "bumblebee.engine.Module.update_wrapper" correctly it can either apply no delay, which means updating every bumblebee-status update, so 1 second (when the interval is unset or set to 0) or 1 or more minutes, according to:
`self._next += int(self.parameter("interval", self._default_interval))*60`
May I know why we can't use a second as a unit?
Without that, I'll be tempted to use "command && sleep 15" without any "shell.interval" and with async set to True to have for example ping results updated every 15s.
Are there any contraindications for changing it "interval" to use seconds (with updating all current modules to use 60x bigger number, except sun.py which is using 3600 which probably was meant as 1h, not 2.5 day ;) )?

tobi-wan-kenobi:
@rrhuffy You are absolutely right. TBH, I'm not sure what I was thinking when writing that code.... I agree that your proposed solution is good, I'm just a bit concerned about backwards compatibility.

I'm thinking about either making the parameter a string, possibly (i.e. "10s" vs. "10m", with minutes being the default), or adding a key/value parameter, so that it's possible to specify either seconds or minutes, and change the existing modules to use the minutes parameter.

I like the latter more, what's your take on that?

rrhuffy:
@tobi-wan-kenobi 
I understand your concerns about backwards compatibility, but after checking the code against "interval" I found README.md + engine.py + only 10 modules.
Going one by one:
- apt is using 30, which seems ok. Half of an hour is ok, half of an minute is too often to check updates.
- currency is using 1 + "currency.interval: Interval in minutes between updates, default is 1.", seems ok.
- dnf is using 30 + "dnf.interval: Time in seconds..." - first module which expects seconds
- github is using 120...seconds: "github.interval: Interval in seconds..." - second module which expects seconds
- getcrypto is using 120...again seconds: "getcrypto.interval: Interval in seconds...", the third one for seconds
- ping is using 60, also seconds: "ping.interval: Time in seconds...". 4th module.
- shell is mine module, but I won't be angry for backwards-compatibility issues :D so count it as the fifth one for seconds ;)
- stock is using 60, without documentation. I believe that in the finance world one hour can be an eternity, so assuming seconds here. 6th module for seconds.
- sun is using 3600. In seconds that is one hour, in minutes it is...2.5 days. Again, this module expects seconds
- weather is using 15 and seems like one quarter is just ok value for checking the weather. 15 seconds is a bit too often ;) So third module using minutes.

So...if I count correctly we have 7 modules which expect seconds versus 3 which expects minutes.
Because of that, I'm very tempted to just forcefully change the interpretation of this field to seconds + change 3 modules using it as minutes to "interval(X*60)" + change shell's documentation + be aware of that change in next few pull requests.
By just changing it to seconds (without doing any changes in modules) you're fixing 70% of modules while breaking 30%. Still on the plus ;)
That long explanation is because I really think that seconds should be the only one unit used. Adding a "600s"/"10m"/"1h" or (15,'s')/(1,'m')/(1,'h') will add unnecessary complexity here and it's a step into the wrong direction for me.

tobi-wan-kenobi:
Good points, thanks for the detailled explanation. TBH, I am mostly concerned about users that have overriden the setting with a parameter. in that case, the setting would be in seconds suddenly.

But I might be able to come up with a solution for that, as well :thinking:

⁣Sent from TypeApp ​

tobi-wan-kenobi:
Hi @rrhuffy !

Since the last commit (621564d), the interval for `shell`, and most of the other modules, should be in seconds.

Thanks a lot for your suggestion!

PR #462, opened 2019-11-10
[modules/cmus] Add cmus.layout property to configure which widgets ar…

tobyp:
This PR adds:
- the `cmus.layout` parameter, which defaults to the space-separated list `cmus.prev cmus.main cmus.next cmus.shuffle cmus.repeat` and can be used to select/arrange the widgets created by the cmus module. Their functionality remains the same.

Currently, the cmus module will always add widgets for the actions "Previous", "Next", "Shuffle", and "Repeat", as well as a "Play/Pause" action on the main widget.
Play, Previous and Next can often be mapped to hardware keys, and Shuffle/Repeat are not needed as often, and probably go together with more complex configuration that requires access to the main cmus screen anyway.
Therefore, to save space in the bar for more of bumblebee's delicious widgets, users can now select which widgets they want on their screen.

### DIFF
--- a/bumblebee/modules/cmus.py
+++ b/bumblebee/modules/cmus.py
@@ -7,6 +7,7 @@ Requires the following executable:
 
 Parameters:
     * cmus.format: Format string for the song information. Tag values can be put in curly brackets (i.e. {artist})
+    * cmus.layout: Space-separated list of widgets to add. Possible widgets are the buttons/toggles cmus.prev, cmus.next, cmus.shuffle and cmus.repeat, and the main display with play/pause function cmus.main.
 """
 
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi! Thanks a lot for the PR, that's a really really nice idea that I'd dearly like to see in `master`.

Unfortunately, the current implementation only works for Python 3.6+, because only there, the standard `dict` inserts keys in order. For older Pythons, the order depends on the map hashing.

Since you don't seem to need the map anywhere, my suggestion would be to switch widget_map to a list to which you simply append. Does that sound like an OK approach to you?

tobyp:
Good catch!

I've updated my PR to use a list. The purpose of the map is to remember input callbacks and register them *after* the widgets are added. This is the order in which the old version did it, but it doesn't seem to matter to the engine/module as far as I can tell. Intuitively I would expect that the widgets have to be known to the engine before callbacks can be registered, so I kept it.

Please let me know your thoughts on this, as I have taken the codeclimate results as a personal challenge and still want to fix them before the weekend, and could simplify this if an ordering requirement is not intended here.

tobi-wan-kenobi:
On first glance, I don't think that widgets needs to be known to the engine before callback registration. As far as I can remember, the widget has a UUID, and that's what's bound to the callback.

tobi-wan-kenobi:
Also: Merging as is, since Travis is fine :) Thanks a lot for the contribution!

tobyp:
Alright, thanks, happy to contribute! Looking forward to the next release :)

PR #463, opened 2019-11-10
Support disabling scrollable by setting width=-1 on the module

tobyp:
This PR adds:
- For scrollable widgets, setting their `width` parameter to -1 makes them resize to their content. 

This can be used to disable scrolling per widget/module.

Auto-size combines well with `cmus.scrolling.makewide=false`.


tobi-wan-kenobi:
Looks like a good and useful addition to me, many thanks!

PR #465, opened 2019-11-12
Quote vpn names when executing nmcli

subutux:
Fixes an issue when VPN names have spaces.


tobi-wan-kenobi:
Ah, good catch, many thanks for the PR!

PR #469, opened 2019-11-24
Use pytz to figure out local timezone dynamically

### LINKED ISSUE #468: [modules/datetime] Wrong timezone after timezone update
#### Summary
Affected module: datetime

#### Description:
After an update of the timezone, bumblebee is stuck to the previous one.

#### How to reproduce

```bash
$ date         
Sun 24 Nov 11:01:20 GMT 2019
$ # Run bumblebee, it shows 11:01
$ sudo ln -sf /usr/share/zoneinfo/America/Chicago /etc/localtime
$ date
Sun 24 Nov 05:01:20 CST 2019
$ # Bumblebee stuck at GMT and still shows 11:01:20
```

I guess this is due to datetime using the timezone acquired at startup in  https://github.com/tobi-wan-kenobi/bumblebee-status/blob/7d989a6fca0bb6bb6c399261ae081b089a44167c/bumblebee/modules/datetime.py#L45.
A solution may come from here https://stackoverflow.com/a/25634136.

cphyc:
Fixes #468.

### DIFF
--- a/bumblebee/modules/datetime.py
+++ b/bumblebee/modules/datetime.py
@@ -15,6 +15,11 @@ from __future__ import absolute_import
 import datetime
 import locale
 import bumblebee.engine
+import os
+try:
+    import pytz
+except ModuleNotFoundError:
+    pytz = None
 
 def default_format(module):
     default = "%x %X"
@@ -41,8 +46,15 @@ class Module(bumblebee.engine.Module):
             locale.setlocale(locale.LC_TIME, ('en_US', 'UTF-8'))
 
     def get_time(self, widget):
+        if pytz:
+            # Get local timezone (see https://github.com/tobi-wan-kenobi/bumblebee-status/issues/468)
+            my_tz_name = '/'.join(os.path.realpath('/etc/localtime').split('/')[-2:])
+            my_tz = pytz.timezone(my_tz_name)
+        else:
+            my_tz = None
+        
         enc = locale.getpreferredencoding()
-        retval = datetime.datetime.now().strftime(self._fmt)
+        retval = datetime.datetime.now(my_tz).strftime(self._fmt)
         if hasattr(retval, "decode"):
             return retval.decode(enc)
         return retval


tobi-wan-kenobi:
Looks good, thanks a lot for reporting the issue and fixing it!

bumblebee/modules/datetime.py
```diff
@@ -15,6 +15,11 @@
 import datetime
 import locale
 import bumblebee.engine
+import os
+try:
+    import pytz
+except ModuleNotFoundError:
```
david-perez: This is Python 3 only and broke my installation of `bumblebee-status` in a system in which `usr/bin/env` points to Python 2. Can we change [this line](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/bumblebee-status#L1) to `/usr/bin/env python3` and stop supporting Python 2 given that it is soon going to be [deprecated tomorrow](https://www.python.org/doc/sunset-python-2/)?

david-perez:
@tehgibbo What error are you getting when you execute `bumblebee-status` manually in your terminal? (i.e. when running the command next to `status_command` in your i3 config).

tehgibbo:
The entire bar crashes with status_error. Starting it in terminal didn't
show any useful information so I had to narrow it down to timedate by
removing modules from the bar config section one by one.

tehgibbo:
$ ./bumblebee-status -d -m datetime

{"version": 1, "click_events": true}
[
[],
  File "./bumblebee-status", line 54, in main
    engine.run()
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/engine.py", line
285, in run
    self.write_output()
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/engine.py", line
300, in write_output
    self._output.draw(widget=widget, module=module, engine=self)
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/output.py", line
130, in draw
    full_text = widget.full_text()
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/output.py", line
104, in full_text
    return self._full_text(self)
  File
"/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/modules/datetime.py",
line 52, in get_time
    my_tz = pytz.timezone(my_tz_name)
  File "/usr/lib64/python3.6/site-packages/pytz/__init__.py", line 180, in
timezone
    raise UnknownTimeZoneError(zone)
pytz.exceptions.UnknownTimeZoneError: 'etc/localtime'

During handling of the above exception, another exception occurred:

  File "./bumblebee-status", line 80, in <module>
    main()
  File "./bumblebee-status", line 74, in main
    output.draw(widget, error)
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/output.py", line
136, in draw
    padding = self._theme.padding(widget)
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/theme.py", line
113, in padding
    return self._get(widget, "padding", "")
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/theme.py", line
248, in _get
    state_themes.append(self._get(widget, state, {}))
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/theme.py", line
266, in _get
    if mod and not mod.parameter("is-unittest"):
  File "/home/cilnnot/GIT-apps/bumblebee-status/bumblebee/engine.py", line
123, in parameter
    value = self._config["config"].get(module_name, default)
TypeError: 'Config' object is not subscriptable




I see the error about /etc/localtime, checked mine and see nothing wrong
with it.

tobi-wan-kenobi:
I guess this happens when /etc/localtime is not a symlink :(

Anyhow, datetime.datetime.now() should be in the correct timezone already, so I am leaning towards reverting the changeset, unless anybody (@cphyc ) objects. I believe restarting the bar after you have manually updated the system time is OK - compared to the bar stopping to work at all.

Or we find a good way to fix this issue :)

david-perez:
@tehgibbo It seems like your timezone is not configured in your system. Try running `ln -s /usr/share/zoneinfo/Europe/Madrid /etc/localtime` (where you should change `Europe/Madrid` to your timezone).

@tobi-wan-kenobi We should keep this changeset only if we want the `datetime` module to be Python 3-only (fair, since it's already being sunset) and if we *require* users to have `/etc/localtime` configured.

We could also manually check if `/etc/localtime` is a symlink, and if it's not, default to `datetime.datetime.now()`.

tehgibbo:
Hi David, the timezone is configured correctly on my gentoo system - by
default (as per install guide) we echo the timezone name into
/etc/timezone, then our package manager runs --config on
sys-libs/timezone-data which copies the timezone to /etc/localtime instead
of symlinking it. I'm sure some other distros behave this way too.

Other programs using my C library have no trouble knowing the timezone with
this setup, as did timedate for 2 years before this code change.

So a simple symlink would 'fix' this for me, however I switched to
datetimetz which works without issue.

david-perez:
@tehgibbo Ah, I see. Interesting how Gentoo does things. I wonder what is the rationale for copying instead of symlinking? Maybe because `/usr` and `/etc` need not be under the same filesystem.
Do you know of any other distros that behave this way?

tobi-wan-kenobi:
@tehgibbo If "datetimetz" works for you, that's great to hear. In that case, I will revert this PR's changes.
Thanks for the notes!

Also: @david-perez Yeah, about Python 2 vs. 3, my gut feeling is that many people will still be using Python 2 for quite some time, so as long as I can support Python 2 without major complications ("unicode handling" *grpmh*), I will try to maintain compatibility.

PR #470, opened 2019-11-24
[Feature] automatic inferrence of local currency

cphyc:
This adds the possibility to infer local currency based on local IP address. This relies on two http requests (https://ipvigilante.com/ to get the local country and https://raw.githubusercontent.com/samayo/country-json/master/src/country-by-currency-code.json to get the mapping from country to currency), so we may want to disable it by default.

A possible improvement would use geoclue for this but I could never figure out how to use dbus...

### DIFF
--- a/bumblebee/modules/currency.py
+++ b/bumblebee/modules/currency.py
@@ -73,10 +73,14 @@ class Module(bumblebee.engine.Module):
         else:
             self._base = src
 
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Looks good to me, thanks a lot!

In general, I'm always a bit hesitant about changing default behaviour, but this particular change makes so much sense to me that I'll just go ahead and merge and see whether somebody complains :)

PR #472, opened 2019-11-27
[modules/nic] Include parameter

xsteadfastx:
After excluding specific network interface prefixes, the include
parameter pulls explicit interfaces back into the interface list.

### DIFF
--- a/bumblebee/modules/nic.py
+++ b/bumblebee/modules/nic.py
@@ -7,6 +7,7 @@ Requires the following python module:
 
 Parameters:
     * nic.exclude: Comma-separated list of interface prefixes to exclude (defaults to "lo,virbr,docker,vboxnet,veth,br")
+    * nic.include: Comma-separated list of interfaces to include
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Hi,

interesting PR, thanks a lot!

Could you please elaborate on the use case of this? I have to admit that I don't get it. From what I see, the same effect could be achieved by just specifying a shorter exclude list that doesn't include anything you put into the "include" list right now, is that correct?

Thanks a lot in advance!

xsteadfastx:
Ok. Maybe you are right i try to explain :)

i have tons of "enx*" interfaces... but i just want one to to be shown in my status bar. So i exclude "enx" and explicit include "enx10653089" for example. Else i would exclude just every single enx interface.

tobi-wan-kenobi:
Ah, in that case, this change makes sense. Thanks for clarifying, will merge.

xsteadfastx:
On Do, 28. Nov 08:11, tobi-wan-kenobi wrote:
>    Ah, in that case, this change makes sense. Thanks for clarifying, will
>    merge.

Oh great :) Thank you!

PR #478, opened 2019-12-11
theme.exclude now takes a comma separated list

JohnDowson:


tobi-wan-kenobi:
Cool addition, thanks!

Looking at the code, shouldn't the code do a split() instead of a strip(), though? Otherwise, `aslist` sometimes returns a list, sometimes a string, or am I missing something?

JohnDowson:
Yeah that's right.

tobi-wan-kenobi:
Looks good, thanks for the quick fix!

PR #480, opened 2019-12-11
Added an option to layout-xkb to only show layout name without a variant

JohnDowson:


tobi-wan-kenobi:
Nice addition, thanks a lot for the PR!

PR #488, opened 2019-12-17
 bumblebee/modules/datetimetz.py

misss13:
Hi I'm using slackware and I had a probelm with date and time cuz status didn't work at all buut datetimetz were working without error but it actually showed [n/a] so I replace it with if so well I write this and it's working =D


tobi-wan-kenobi:
Hi,

thanks for the PR!

Looking at the diff, it seems to comprise other modules besides datetimetz.py as well. Would it be possible to rebase the changesets you actually want to go have in the main release onto master?

Actually, upon a closer look, datetimetz.py itself is *not* part of the list of affected files. Is it possible you accidentially opened a PR from the wrong branch?

tobi-wan-kenobi:
I'll close this PR, please feel free to open a new one with the actual changeset, many thanks in advance!

PR #489, opened 2019-12-18
Bluetooth module: dbus object path and popup menu

### LINKED ISSUE #483: bluetooth: no interface “org.blueman.Mechanism” on object at path /
#### Summary
Affected module: bluetooth

The right-click method is calling a D-Bus method at the wrong D-Bus object path.

#### Description:
(Disclaimer: I'm not a system programmer, so the language and terms I use may be incorrect ...)

The behavior of a right-click at the bluetooth widget is to toggle the bluetooth. This is done with calling a blueman method through D-Bus by using `dbus-send` command, see line [98-100](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/59ad411b2f04ff9916bc84d3cd207550465f7c03/bumblebee/modules/bluetooth.py#L98-L100). The problem I am having now is that, the `org.blueman.Mechanism` interface is not at path `/` on my system. Instead, it's at `/org/blueman/mechanism`. So I have to modify the code to 

```
        cmd = "dbus-send --system --print-reply --dest=org.blueman.Mechanism"\
              " /org/blueman/mechanism org.blueman.Mechanism.SetRfkillState"\
              " boolean:{}".format(state)
```
to make it work.

Not sure if my case is a special case or not. My system is:
* Archlinux, kernel 5.4.2-arch1-1
* Blueman 2.1.1-2
* Not sure if this matters, but I'm using Sway WM and swaybar + bumblebee-status

#### How to reproduce

1. Include the `bluetooth` module in `bumblebee-status`
2. Right click the bluetooth widget
3. See if the right-click toggled the bluetooth on/off

piyueh:
This PR has two modifications in the module ***bluetooth***:

1. Fix #483 
  The cause of #483 is the path to blueman D-Bus object. By default, the path is hard-coded to `/`. This PR adds a parameter to configure this path. The parameter added is `bluetooth.dbus_destination_path`. And the default value is `/`.

2. Allow users to disable popup menu
  The default behavior of a right-click is to pop up a graphical menu. But the graphical menu does not work well in [Sway WM](https://github.com/swaywm/sway) and swaybar because of the Wayland backend. So this PR adds a parameter to control whether to use the graphical menu or not. If `bluetooth.right_click_popup` is `True` (default), a menu pops up, just like before. If `bluetooth.right_click_popup=False`, a right-click will simply turn on/off the bluetooth device without any menu.

### DIFF
--- a/bumblebee/modules/bluetooth.py
+++ b/bumblebee/modules/bluetooth.py
@@ -5,6 +5,7 @@ Parameters:
     * bluetooth.device : the device to read state from (default is hci0)
     * bluetooth.manager : application to launch on click (blueman-manager)
     * bluetooth.dbus_destination : dbus destination (defaults to org.blueman.Mechanism)
+    * bluetooth.dbus_destination_path : dbus destination path (defaults to /)
 
 """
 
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good to me, many thanks for the PR!

PR #490, opened 2019-12-19
Improve battery-upower

piyueh:
1. Add icons for **battery-upower**. The icons are the same set used by **battery** and **battert_all** modules.
2. Fix the parameter names in documentation
3. Fix the issue that when using `bumblebee-status -l modules`, the doc of **arch-update** and **battery-upower** are not separated by a bland line.
4. Now also show the remaining time to fully charged when charging
5. Reduce the frequency of making requests to D-Bus:
  When requesting the remaining time, we only need to request the state and time to empty (or time to full). Original code called a function `get_full_device_information`, which makes a lot of unnecessary requests to get full status. This PR eliminates this behavior by only requesting state and time to empty/time to full.

### DIFF
--- a/bumblebee/modules/arch-update.py
+++ b/bumblebee/modules/arch-update.py
@@ -1,4 +1,5 @@
-"""Check updates to Arch Linux."""
+"""Check updates to Arch Linux.
+"""
 
 
 import subprocess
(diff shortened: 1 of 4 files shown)


tobi-wan-kenobi:
Looking good, thanks a lot for the improvement of batter-upower!

PR #496, opened 2019-12-24
Fixed typo for city output def

CPrompt:


tobi-wan-kenobi:
Good catch, many thanks!

PR #497, opened 2019-12-25
hddtemp module

somospocos:
The docstring of the module should be self-explanatory.

Note: I had to use an ugly hack - catch AttributeError in _fetch_data(), otherwise instantiating Module in TestHddtempModule.setUp() fails. I have zero experience with using mock, so I didn't succeed with mocking socket.socket to avoid this problem. You are very welcome to do this the right way if this module is worth merging.

### DIFF
--- a/bumblebee/modules/hddtemp.py
+++ b/bumblebee/modules/hddtemp.py
@@ -1,0 +1,90 @@
+# -*- coding: utf-8 -*-
+
+"""Fetch hard drive temeperature data from a hddtemp daemon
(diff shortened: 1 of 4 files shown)


tobi-wan-kenobi:

somospocos:
There's something wrong with the table layout on https://github.com/tobi-wan-kenobi/bumblebee-status/wiki/Available-Modules

The hddtemp module description takes two rows.

tobi-wan-kenobi:
Fixed, thanks for raising!

PR #499, opened 2019-12-26
supported more tags

P-Storm:
More MPD tags support. First time using Python so feel free to comment on my code :)

### DIFF
--- a/bumblebee/modules/mpd.py
+++ b/bumblebee/modules/mpd.py
@@ -79,7 +79,27 @@ class Module(bumblebee.engine.Module):
     def _load_song(self):
         info = ""
         try:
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good to me, thanks for the contribution!

Except the "tags" dict should actually be an array ([]), no? I'll merge the PR and update that manually, please let me know if that is a mistake on my side.

P-Storm:
I was looking for a 'set' collection, so no multiple same strings could be in the collection. Did I make it a dictionary on mistake?

tobi-wan-kenobi:
Oh, sorry, you are right! I will revert my changes again as soon as possible.

Thanks again for the PR!

⁣

P-Storm:
@tobi-wan-kenobi 
Should the Syntax for these items be updates in the Available Modules list under Notes? Like the link 'https://www.mankier.com/1/mpc'

PR #500, opened 2019-12-26
Added filtering in modules

P-Storm:
This filters the modules list

An bug that I couldn't solve without changing the argument is that the help file is a bit less descriptive. Not sure how I could change argparse to make this better.

Old:
```
 -l {modules,themes}, --list {modules,themes}
                        Display a list of either available themes or available
                        modules along with their parameters.
```
New:

```
 -l LIST [LIST ...], --list LIST [LIST ...]
                        Display a list of either available themes or available
                        modules along with their parameters.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,12 @@ $ ./bumblebee-status -l themes
 $ ./bumblebee-status -l modules
 ```
 
+You can also narrow you modules list using the filter at the end:
+```
+$ cd bumblebee-status
+$ ./bumblebee-status -l modules xkcd zpool
+```
+
 Any parameter you can specify with `-p <name>=<value>`, you can alternatively specify in `~/.bumblebee-status.conf` or `~/.config/bumblebee-status.conf`. This parameters act as a **fallback**, so values specified with `-p` have priority.
 
 Parameters can also be used to override theme settings, such as:
--- a/bumblebee/config.py
+++ b/bumblebee/config.py
@@ -24,26 +24,29 @@ class print_usage(argparse.Action):
         self._indent = " "*2
 
     def __call__(self, parser, namespace, value, option_string=None):
-        if value == "modules":
-            self.print_modules()
-        elif value == "themes":
+        if value[0] == "modules":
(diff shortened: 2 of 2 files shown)


tobi-wan-kenobi:
I see what you mean. I like the idea, but I agree that it is unfortunate that the help text becomes more generic.

I have two ideas on how to fix this:
* Add a separate "filter" argument that can optionally be specified (i.e. -l modules -f <filters>)
* Make a separate --show-theme and --show-module argument that has a mandatory parameter (list of themes/modules to show)

Do any of those sound acceptable to you?

P-Storm:
> 
> 
> I see what you mean. I like the idea, but I agree that it is unfortunate that the help text becomes more generic.
> 
> I have two ideas on how to fix this:
> 
>     * Add a separate "filter" argument that can optionally be specified (i.e. -l modules -f )
> 
>     * Make a separate --show-theme and --show-module argument that has a mandatory parameter (list of themes/modules to show)
> 
> 
> Do any of those sound acceptable to you?

My question is, should the command be backword compatable? Otherwise I could implement something like:

./bumblebee-status list -theme -filter a b c

with the subcommand list you can get the following:
* -n name "name"
* -d Display (generate Markdown/ show parameters info / show all): default is normal
* -l modules : same as the old one

This way you still keep the old querying, but creating a new structure for other outputting.

Still it is optional to remove the old command (or show in the output that the command is renamed to a new command)

tobi-wan-kenobi:
If possible, I'd like to maintain backwards compatibility.

Having said that, I think implementing a "show-modules" and "show-themes" command in terms of what you suggest (using a subcommand) makes a lot of sense. That way, the old code would stay untouched, and there'd be a separate way to show details about a single module.

tobi-wan-kenobi:
Closing for now, due to inactivity.

Having said that, I am really looking forward to those changes, so if you have something, please feel free to open another PR. If I have time, I will look into it myself (though, admittedly, that's unlikely right now).

PR #504, opened 2019-12-27
Mpd module improvement

### LINKED ISSUE #482: Bug in mpd module?
#### Summary
Affected module: mpd

#### Description:
I am posting this as question because I don't fully understand the code yet.

My use case is that my music files don't have any tags, they are all in **"artist - song name.ogg"** format. So I went to read the source file mpd.py to see what config options do I have and if it's possible to tune the plugin to display what I want.

In the docstring of that file I read about **mpd.format**. For a start, I change it to **"{file}"**. Doesn't work. The first problem is that the default value of mpd.format is not documented properly in the docstring. I scroll down and see:

    self._fmt = self.parameter("format", "{artist} - {title} {position}/{duration}")

at line 52.

So I was probably wrong. Second attempt is to try with **mpd.format="{file} {position}/{duration}"**. This does not work either. I start reading the code further.

    info = bumblebee.util.execute('mpc -f "tag artist %artist%\ntag title %title%"' + self._hostcmd)

this is line 82 and to me it really looks like the **"{artist} - {title}"** is simply hardcoded here into the mpc call, while the docstring advertises I can change it (see above) - at least this is how I understand it.

So is this a bug in the mpd module or am I doing anything wrong?

This is just the first half of my problem. The second one is that when I run **mpc -f "%file%"**, I get a string that looks like **"path/relative/to/mpd/music/directory/root/artist - song name.ogg"**. So even if **"{file}"** would give me what I want, I would still need to post-process the string:

- to remove the ".ogg" extension
- to split it by "/"
- to get the last item which would be "artist - song name" that will get displayed

I'm afraid I don't know how to do this inside the mpd.format option.

Considering all of this, perhaps it would be a better idea for me to just fork this module and hack it to work properly for my particular case?

#### How to reproduce

Described above.

somospocos:
I have documented the previously new added tags in the module docstring, so they show up in the output of '-l modules'.

I have also added two more tags related to the name of file currently being played.

If this is getting merged, I think it closes #482

### DIFF
--- a/bumblebee/modules/mpd.py
+++ b/bumblebee/modules/mpd.py
@@ -7,7 +7,31 @@ Requires the following executable:
     * mpc
 
 Parameters:
-    * mpd.format: Format string for the song information. Tag values can be put in curly brackets (i.e. {artist})
+    * mpd.format: Format string for the song information.
+      Supported tags (see `man mpc` for additional information)
+        * {name}
+        * {artist}
+        * {album}
+        * {albumartist}
+        * {comment}
+        * {composer}
+        * {date}
+        * {originaldate}
+        * {disc}
+        * {genre}
+        * {performer}
+        * {title}
+        * {track}
+        * {time}
+        * {file}
+        * {id}
+        * {prio}
+        * {mtime}
+        * {mdate}
+      Additional tags:
+        * {position} - position of currently playing song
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good, many thanks!

PR #506, opened 2019-12-29
Use $SHELL in the shell module

KrzysiekJ:


tobi-wan-kenobi:
Hm, I assumed subprocess would use the default shell configured on the system.

Still, I am prepared to merge the PR, but only if you add security guards that retain backwards compabitility if os.environ['SHELL'] is not set for some reason - right now, the module would fail.

KrzysiekJ:
Indeed. I thought that `$SHELL` is mandatory, as it is defined in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/), but apparently it is not the case:

> If the variables in the following two sections are present in the environment during the execution of an application or utility, they shall be given the meaning described below.

I too was surprised by the fact that Python does not by default use the preferred shell, but maybe this is for the sake of interoperability…

Updated the PR.

tobi-wan-kenobi:
Yeah, I found out that Python just uses /usr/bin/sh, so doesn't take per-user settings into account. Weird.

I'm sorry, but I don't see any updates to the PR. I see you pushed something else, but the changeset still shows only the original changes.

KrzysiekJ:
It is updated. `os.environ['SHELL']` was changed to `os.environ.get('SHELL')`. Therefore if `$SHELL` is not present, the argument falls back to the default value of `None`.

tobi-wan-kenobi:
Ah, got it, I overlooked that. Thanks for clarifying!

PR #508, opened 2019-12-30
layout-xkbswitch module

somospocos:
Another layout module that relies on xkb-switch (https://github.com/ierton/xkb-switch).

Shortest size with pretty much same functionality, because it doesn't have any heavy output parsing: xkb-switch is very laconic.


tobi-wan-kenobi:
Looks good to me, many thanks for the new module! Hopefully, now there's a solution for *every* keyboard switching taste :)

PR #509, opened 2019-12-30
Added simple yubikey state module

EmmaTinten:
A simple module which shows if a yubikey is present (displaying serial number) or not.


tobi-wan-kenobi:
As a Yubikey user myself: Great addition, many thanks!

PR #516, opened 2020-01-02
[modules/cmus] add file tags

somospocos:
Improvement similar to 1238914f and fac363c9 for mpd module: add file tags.


tobi-wan-kenobi:
Looks good, many thanks!

PR #519, opened 2020-01-04
[modules/mpd] refactor to use layout

somospocos:
Refactored mpd module to use layout for widgets instead of hardcoding them - the way cmus module already does it.


tobi-wan-kenobi:
Looks good to me, many thanks!

PR #522, opened 2020-01-07
[modules/pulseaudio] add support for volume bars

somospocos:
Added support for volume bars next to numerical values in percents. Bars are not added as an additional module widget, to save a bit of space.

The change should be backwards compatible, because bars are switched off by default.

bars


tobi-wan-kenobi:
Very nice, thanks!

PR #523, opened 2020-01-10
added option for displaying the power consumption

hoamer:
hi,

I added an option to display also the actual power-consumption.
It can be enabled via "battery.showpowerconsumption=true".

It is just a quick implemented feature, so please review it if you have time ;)

best regards
Sebastian

### DIFF
--- a/bumblebee/modules/battery.py
+++ b/bumblebee/modules/battery.py
@@ -79,17 +79,24 @@ class Module(bumblebee.engine.Module):
                 capacity = int(f.read())
         except IOError:
             return "n/a"
+
         capacity = capacity if capacity < 100 else 100
         widget.set("capacity", capacity)
-        output =  "{}%".format(capacity)
+
+        # Read power conumption
+        if bumblebee.util.asbool(self.parameter("showpowerconsumption", False)):
+            r=open(widget.name + '/power_now', "r")
+            output =  "{}% ({})".format(capacity,str(int(r.read())/1000000) + "W")
+        else:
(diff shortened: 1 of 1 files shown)


bumblebee/modules/battery.py
```diff
@@ -79,17 +79,24 @@ def capacity(self, widget):
...
+
         widget.set("theme.minwidth", "100%")
-        
         if bumblebee.util.asbool(self.parameter("showremaining", True))\
                 and self.getCharge(widget) == "Discharging":
             output = "{} {}".format(output, self.remaining())
 
         if bumblebee.util.asbool(self.parameter("showdevice", False)):
-            output = "{} ({})".format(output, os.path.basename(widget.name))
+            output = "{} ({})".format(output, r.read())
```
tobi-wan-kenobi: That is dangerous, isn't it? "r" might not be initialized (because it only is when "showpowerconsumption" is set.

tobi-wan-kenobi:
Thanks, that's a nice addition. Looking at the code, I spotted one problem that seems to break functionality if "showpowerconsumption" is set to false, can you please verify and fix that?

Once that is done, I'll go ahead and merge, many thanks in advance!

hoamer:
Hi,

sorry for the late response.
@tobi-wan-kenobi yes, the line was unintentionally changed. I fixed it.

I am sorry, but I can not reproduce a problem when setting "battery.showpowerconsumption" to false.
How does it affect it negatively on your site?

tobi-wan-kenobi:

PR #525, opened 2020-01-12
bumblebee.util.bytefmt() refactor

somospocos:
I am planning to add a format parameter to **traffic** module, but it uses **bumblebee.util.bytefmt()**, which hardcodes the format. So I've refactored that function and added a format argument. It has a default value equal to the hardcoded one, so it should be backwards compatible.


tobi-wan-kenobi:
Looks good, thanks for the improvement!

PR #527, opened 2020-01-13
[modules/traffic] add format parameter

somospocos:
There is a small problem with this change, which I don't know how to deal with.

Previous improvement of **bumblebee.util.bytefmt()** and this change were made because I want to have shorter (integer) values displayed by traffic module.

I intend to use **-p traffic.format="{:.0f}"**. But this leaves empty gaps in the bar for short values. The offending code line is:

    widget = self.create_widget(widgets, name, attributes={"theme.minwidth": "1000.00MB"})

What is the purpose of **theme.minwidth** here? If I remove this hardcoded value, everything is fine - the formatted text takes the minimum amount of space it needs. But I didn't want to just remove it without understanding its purpose and the implications of it.


tobi-wan-kenobi:
The "minwidth" helps i3bar figure out the minimum width of a widget. Hence, in the code above, you'd need to take the custom format specifier in account, as well.

tobi-wan-kenobi:
Great, thanks a lot!

Issue #505, opened 2019-12-29
Towards diversifying output options


somospocos:
I have played a bit with the approach that I described here https://github.com/tobi-wan-kenobi/bumblebee-status/issues/486#issuecomment-568448248 and I've ended with https://github.com/somospocos/pytxtbar

I wonder if you mind merging this into bumblebee/output.py to expose the **hbar()** and **vbar()** functions to module writers.

tobi-wan-kenobi:
Not 100% sure how that would look like in detail, but I am certainly open to a PR.

Looking at the demo script output, I am thinking this might make more sense as a module ("graph.py", "chart.py", something like that) that a module could use to produce output (somewhat similar to how popup.py offers functionality to produce popups?

somospocos:
popup.py seems to be using tkinter - that's a whole third party GUI library in contrast with the pure CLI app that bumblebee-status is, so having that kind of stuff in its own module is perfectly justified.

While my demo is simply something like "visual representation of a number in ascii art". I see hbar() and vbar() just as a complement to scrollable() function from output.py (which is just technically "animated text in ascii art").

Or you just mean having the code separated in its own module and then just exposing hbar() and vbar() as functions in output.py?

Technically, I've just tried to implement something similar to gdbar (https://linux.die.net/man/1/dzen2-gdbar) from dzen2.

somospocos:
No pull request yet, but I've got something to show. This is how I imagine it - see https://github.com/somospocos/bumblebee-status/tree/txtbar.

And a couple of trivial examples in https://github.com/somospocos/bumblebee-status/tree/txtbar-examples - just diff the modules pulseaudio.py / pulseaudio1.py / pulseaudio2.py to see the usage and use them together like this:

    bumblebee-status -m pasink1 pasink2 -t gruvbox-powerline

If it gets green light - I will make a pull request for the txtbar branch.

tobi-wan-kenobi:
Pretty cool, I'm looking forward a lot to having this as a PR!

The only thing: It might be nice to tie the actual characters to be used into the icon sets (themes/icons/), instead of hardcoding them, but that can also be done later.

somospocos:
> It might be nice to tie the actual characters to be used into the icon sets (themes/icons/), instead of hardcoding them

[...]

PR #528, opened 2020-01-19
[core/output] add support for drawing graphs using Braille chars

somospocos:
As requested here https://github.com/tobi-wan-kenobi/bumblebee-status/issues/505#issuecomment-573737304


tobi-wan-kenobi:
Looks very nice, many thanks for the contribution!

PR #539, opened 2020-01-31
python packaging and allowing for pip install usage

ginsburgnm:
Hey, so this:

  * adds versioneer -> every time you git-tag a release the version is updated based on the git tag
  * adds github actions -> every time you do a release it gets uploaded to pypi
    * you'll need to setup the pypi username/secret for your remote inside your repository settings
  * adds ability for a user that pip installed bumblebee status to run bumblebee status from their python environment, currently all your setup script installs is the library, not the actual main script itself
  * changes the setup to utilize the setup.cfg file, which is more standard than using an __about__.py somewhere in your repo
  * cleans up setup.cfg a lot
  * packages up the themes and icons (they are now in the pip package, but not yet used by the codebase)
    * what is still needing to be done before this is truly pip installable and usable is changing how the themes/icons are accessed to using [pkg_resources](https://kite.com/python/docs/pkg_resources) 

Hope you find this useful.


tobi-wan-kenobi:
Very cool, thanks! I especially like the github workflow.

TBH, I don't fully get why versioneer.py is included in its entirety, but I trust you know what you are doing :)

I'll merge this for now, investigate pkg_resources and will also try to change the github workflow to use API keys.

Thanks for your efforts!

tobi-wan-kenobi:
@ginsburgnm I freely admit I am too dumb to understand `pkg_resources`, or how the whole packaging works, at all.

I've played around a bit now - mainly my goal is to have a PyPI package that people can actually use - and the only "solution" I found was to include the icons and themes as "data_files" in setup.py and extend the theme lookup logic to relatively search also for those in the pip (which is really ugly).

Any insights on how to improve that would be much appreciated!

ginsburgnm:
@tobi-wan-kenobi new PR in; it's honestly about what you were saying (If I read what you said correctly) 

> I freely admit I am too dumb to understand pkg_resources, or how the whole packaging works, at all.

We all are :smile: I have like, one python repo that works as I expect, and I basically just copy-pasta the stuff I need to bootstrap new ones. That's the dirty secret of it all, none of us *actually* know what we're doing. The fact that we publish our code publicly continues to amaze me, it's like having a teacher look over your shoulder while you take a test but worse. :laughing:

PR #540, opened 2020-02-05
using pkg_resources

ginsburgnm:
Hey, saw your request for further help.

Due to how you structured `theme.py` it seems pretty straightfoward; all that really appears to be necessary is adding the additional directory check for pkg_resources. I did need to move `themes` up a directory for pkg_resources to appropriately see it though. for backwards compatibility I created a symlink from the original location to the new one.

I also did a bit of cleanup based on what my linter showed. feel free to pick out only the stuff you want (or if you completely hate it then scrap the whole thing, it's your repo lol); I tested locally (which as we all know is obviously all the necessary testing :roll_eyes: ) with a fresh virtualenv run from a completely separate location and the theme looked right. So hopefully it'll work on a real pip install.


tobi-wan-kenobi:
Awesome! Many thanks. I will carefully review the PR over the weekend, so far, the changes look totally fine to me


Not sure I like the move of themes, though. It feels ... redundant to have all project files in a subdirectory of a git repo, just to please packaging... But pkg_resources wants what pkg_resources wants, no?

Also thanks for your comments in the other PR, good to hear I'm not the only one :-P
⁣

tobi-wan-kenobi:
So, looking at the PR in some detail:
1. I learnt some new fun things about Python - thanks!
2. Most of the changes look really good to me, and are definitely improvements
3. However, that the themes need to be moved, I don't like - I am particularly worried about how that will affect other packages (I know at least about one AUR package and one SuSE package I'm not maintaining myself) - do you think it might be an option to reverse the solution and add the symlinks for the pkg_resources?

Thanks a lot for your efforts!

ginsburgnm:
@tobi-wan-kenobi  I've seen some funky behavior with symlinks in python packaging, but I'll take a look and we'll see what we can do

ginsburgnm:
(locally speaking) reversing the symlinks and changing the MANIFEST seems to have worked. I had a temporary theme only residing in the git repo after my packaging was done, tried to that after install and it failed, then tried to load in `dracula-powerline` and it was able to find that one. So it wasn't looking at the git repo but at it's package resources.

I believe this is what we're looking for @tobi-wan-kenobi

tobi-wan-kenobi:
Great, many thanks!

The diff looks totally fine for me, will merge it and see who complains :)

ginsburgnm:
@tobi-wan-kenobi completely removed bumblebee-status repo from my laptop; did a pip install from master (`pip install git+https://github.com/tobi-wan-kenobi/bumblebee-status.git@master`), changed where to look for bumblebee-status but nothing else in my i3config; the theme, icons and modules all loaded up properly! I think we're good!

tobi-wan-kenobi:
Many thanks for the careful testing!

Your changes should be in master already.

Issue #531, opened 2020-01-23
End user needs more direct low level control over pango markup of prefix/suffix icons


somospocos:
I am "guilty" of lobbying for pango markup support (issue #493) and custom prefix/suffix colors (issue #520) and I've just bumped into a situation which proves that the quick hacks bumblebee has got for making me happy, might not be enough in the long term.

I took a good beating from the technical debt, so hopefully this issue will influence the ongoing redesign of bumblebee in the development branch.

While bumblebee supports FontAwesome for icons, this might not be enough for a capricious and picky user (like me) for whatever reasons, technical or aesthetical ones.

I have been toying with Nerd patched fonts (https://www.nerdfonts.com/) which bundle FontAwesome icons and some other icon packs. I did so by making a copy of **awesome-fonts.json**, renaming it to **nerd-fonts.json** and changing things there, then using **-i nerd-fonts** when invoking bumblebee-status.

I have been experimenting with a narrow font (Iosevka) and a smallish font size, for the sake of saving screen space and packing as much info as possible in a single status line. I have noticed that for narrow fonts, the icons are **very** small, way smaller than providing a good visual experience.

I was able to come up with a workaround, but... not in context of bumblebee-status.

This is what I've been playing with:

The Nerd patched font Iosevka that can be downloaded from here https://github.com/ryanoasis/nerd-fonts/releases/download/v2.0.0/Iosevka.zip

    bar {
        id top
        workspace_buttons no
        font pango:Iosevka 9
        position top
        status_command test.sh
    }

this is the relevant snippet from the i3 config file.

And this is the status command script:

    #!/bin/sh

    echo '{"version": 1}'
    echo '['
    echo '[]'

    while true; do
        echo ',[{"full_text": "tiny icon", "markup": "pango"}, {"full_text": "", "markup": "pango"}, {"full_text": "some text", "markup": "pango"}, {"full_text": "adjusted size icon", "markup": "pango"}, {"full_text": "<span font_desc=\"Iosevka 16\"></span>", "markup": "pango"}, {"full_text": "some text", "markup": "pango"}, {"full_text": "adjusted size + adjusted offset icon", "markup": "pango"}, {"full_text": "<span font_desc=\"Iosevka 16\" rise=\"7500\"></span>", "markup": "pango"}, {"full_text": "some text", "markup": "pango"}]'
        sleep 1
    done

[...]

PR #549, opened 2020-02-12
Iconmarkup

somospocos:
Implementation of low level access to Pango markup discussed here https://github.com/tobi-wan-kenobi/bumblebee-status/issues/531#issuecomment-578405381

For

    font pango:Iosevka 9

in i3 config, this is how mpd module looks by default:

default

and this is how it looks with

    --iconmarkup="<span font_desc='Iosevka 15' rise='-3500'>{}</span>"

iconmarkup

I am quite pleased with the visual result.

### DIFF
--- a/bumblebee/config.py
+++ b/bumblebee/config.py
@@ -17,6 +17,7 @@ THEME_HELP = "Specify the theme to use for drawing modules"
 PARAMETER_HELP = "Provide configuration parameters in the form of <module>.<key>=<value>"
 LIST_HELP = "Display a list of either available themes or available modules along with their parameters."
 DEBUG_HELP = "Enable debug log, This will create '~/bumblebee-status-debug.log' by default, can be changed with the '-f' option"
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Looks good to me, "xml" is in the Python Standard Library, so that is fine. Also, I really like how you're merging the theme fg/bg colors, if none are specified.

somospocos:
Initially I considered undoing the change for support of prefix foreground/background in themes that you added earlier at my request, especially because no theme that officially ships with bumblebee-status makes use of these settings yet, and because those colors can be set directly in the pango markup template now, but I decided to rather add some new functionality on top of already existing ones for a few reasons:

- neglecting backwards compatibility could eventually break some privately used third party themes that actually make use of prefixfg/prefixbg, thus forcing users to switch to this new way
- now users simply have multiple choices: if they are happy with a default theme, they just use it, if they want just custom colors for icons, they can copy the default theme to ~/.config/bumblebee-status, add the prefixfg/prefixbg things and be happy, if they encounter additional problems like me with those tiny icons when I switched to Nerd fonts (which are not officially supported yet, but hopefully this new feature makes that closer) - they can resort to this lowest level of customization and manually hack the pango markup template for icons until the desired result is achieved
- hopefully this diversity of options for the end user can help you get a better perspective on what the end user needs and influence your design decisions for the development branch
- as I wrote earlier, initially adding yet another layer of complexity on top of existing drawing logic looked troublesome, so I'm glad I took time to split that code into smaller parts, which ended being easier to work with

PR #560, opened 2020-02-20
Remove print statement when aligning text

### LINKED ISSUE #550: Scrollable spotify is left aligned
#### Summary
Affected module: spotify

#### Description:
Thanks to #451 longer text is now scrolled, however if the title is shorter than the maximum allowed, it looks as if it floats in the middle of nothing. I can see the logic behind, but it does look weird I think, to me at least.

Possible solutions:
1. make text (and icon) right-aligned (or make it configurable) - this in essence was the behavior before
1. make scrollability configurable (enabled/disabled)

I think 1st option would be best.

#### How to reproduce
Screenshot_2020-02-12_20-40-34

ljupchokotev:
Fixes #550. The print statement results in a "Could not parse JSON" error.

### DIFF
--- a/bumblebee/output.py
+++ b/bumblebee/output.py
@@ -74,7 +74,6 @@ def scrollable(func):
         if len(text) <= width:
             # do alignment
             align = module.parameter("theme.align", "left")
-            print("alignment: {}".format(align))
             if align == "right":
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
That's an embarassing one - thanks for the fix!

PR #562, opened 2020-02-26
Added modules/libvirtvms.py but no way to unit test.

maxpivo:
Would like to add a module for displaying active libvirt VMs. Do you have instructions on how to unit test these modules? The code works outside of bumblebee status but not within it, so not sure where to do from here. Love bumblebee-status btw, been using for several years now.


somospocos:
There are some unit test examples in the repository, I believe the general approach is to use mocks where you have to test code coupled to external resources. Me still being a mock-illiterate, I can't go into details about how to do it the right way.

For a start, I would try to run bumblebee in debug mode, see -d and -f arguments.

By the way, it could be just my web browser's problem, but your code appears wrongly indented at lines 5-6 and 24-end of file.

tobi-wan-kenobi:
Thanks for the addition!

I agree with @somospocos, debug mode is probably your best bet. For unit testing, you'd have to mock out the libvirt stuff, which is doable, but not sure what this would bring.

Especially when it's about syntax errors, unittests are notoriously unhelpful :)

tobi-wan-kenobi:
Looks good, many thanks!

PR #563, opened 2020-02-27
import popup_v2

naota:
The vpn module is using bumblebee.popup_v2, but does not import it, causing nothing on click.


tobi-wan-kenobi:
Good catch, thanks!

PR #564, opened 2020-02-27
Properly connect to the selected VPN profile

naota:
The vpn module is trying to connect "self._connected_vpn_profile", but it is None when no VPN is connected. As a result, the VPN connecting menu is not working. Use "self._selected_vpn_profile" to connect the selected one.


naota:
BTW, do we really need self._selected_vpn_profile in this module? Maybe, we can just drop this variable and use the "name" variable here.

tobi-wan-kenobi:
Not sure about the latter, @bbernhard ?

Other than that: Thanks for the PR!

PR #565, opened 2020-02-28
[modules/cpu2] align cpu load

somospocos:
This fixes a rough edge in cpu2 module where variable size of total cpu
load value string was making the whole bar slide during updates. CPU
load is right-aligned now and takes 3 chars in the widget, so load can
go up to 100 and still fit.


tobi-wan-kenobi:
Looks good. BTW: You should also be able to use the "theme.minwidth" to achieve the same goal (see cpu.py for an example)

PR #569, opened 2020-03-05
modules/traffic] Remove debug print()

somospocos:
fixes 6171c792


tobi-wan-kenobi:
Thanks for fixing this!

PR #573, opened 2020-03-17
[module] improvement of docker_ps by adding states

martindoublem:
Added state to the module to indicate when no containers are running, vs daemon off, container running


tobi-wan-kenobi:
Looks good to me, thanks for the addition!

Minor note: For efficiency, you could cache the result of the status() call in the widget itself (widgets have get/set methods for key/value pairs), and could then retrieve that to determine the status, instead of having to call status() twice a second, essentially.

PR #575, opened 2020-03-18
[module] New module bluetooth2

martindoublem:
I was unhappy with how the module bluetooth functioned so I decided to write a new one

This is the result. The module tries to find adapters ON with bluetoothctl list initally
if none are there, then it assumes it is off and check if any adapters are present with rfkill list

State were added

as well as the number of devices

I also made it so that the toggle works

I was initally going to push this as bluetooth but decided it was so different so to push it as a different module.

python-dbus is a dependency


tobi-wan-kenobi:
Looks good to me, many thanks for the addition!

PR #578, opened 2020-03-21
[module/cmus] Add server and passwd parameters

TheEdgeOfRage:
Currently the cmus module tries to connect to the local cmus unix socket, but when cmus is started with the --listen option, that socket isn't openend.

I've added two parameters, (server and passwd) to allow bumblebee to connect to cmus through a TCP socket, be it remote or local.


tobi-wan-kenobi:
Glad to see you back :)

Diff looks good, in it goes, thanks for the contribution, as always!

PR #580, opened 2020-03-26
Requirements of sun module

soykan:
I append the package requirements as comments. 
It will show on `bumblebee-status -l modules`


tobi-wan-kenobi:
Thanks a lot! Improvements to the documentation are always welcome :)

PR #582, opened 2020-03-26
[module] sensors2 made selective

martindoublem:
made the module sensors2 selective by adding the following parameters and their handling:

    * sensors2.chip_include: Comma-separated list of chip to include (defaults to "" will include all by default, example: "coretemp,bat")
    * sensors2.chip_exlude:Comma separated list of chip to exclude (defaults to "" will include none by default)
    * sensors2.field_include: Comma separated list of chip to include (defaults to "" will include all by default, example: "temp,fan")
    * sensors2.field_exlude: Comma separated list of chip to exclude (defaults to "" will exclude none by default)
    * sensors2.chip_field_exclude: Comma separated list of chip field to exclude (defaults to "" will exclude none by default, example: "coretemp-isa-0000.temp1,coretemp-isa-0000.fan1")
    * sensors2.chip_field_include: Comma-separated list of adaper field to include (defaults to "" will include all by default)

The chip, field are findable by using sensors -u

### DIFF
--- a/bumblebee/modules/sensors2.py
+++ b/bumblebee/modules/sensors2.py
@@ -10,6 +10,12 @@ Parameters:
     * sensors2.showfan: Enable or disable fan display (default: true)
     * sensors2.showother: Enable or display "other" sensor readings (default: false)
     * sensors2.showname: Enable or disable show of sensor name (default: false)
+    * sensors2.chip_include: Comma-separated list of chip to include (defaults to "" will include all by default, example: "coretemp,bat")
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks great to me, many thanks for this addition!

martindoublem:
Yeah, I saw  that some requested a few months back and I thought I could do so, in the end it makes this module much more usuable than its previous form; I used to create a sensors3 on my machine to make it keep just what I wanted, now I can just filter it out.

PR #585, opened 2020-03-28
[module] New module SMART status checker fo hdd/ssd

martindoublem:
Smart status checker for hdd/ssd

Paramaters:
	* smartstatus.display: how to display (defaults to "combined", other choices: "seperate" or "singles")
        * smartstauts.drives: in the case of singles which drives to display, separated comma list value, multiple accepted (defaults to "sda", example:"sda,sdc")

requires: smartmontools

smartcl should be given sudo access with no password for this to work
(per example adding: %wheel ALL=(ALL) NOPASSWD: /usr/bin/smartctl)

### DIFF
--- a/bumblebee/modules/smartstatus.py
+++ b/bumblebee/modules/smartstatus.py
@@ -1,0 +1,117 @@
+# -*- coding: UTF-8 -*-
+
+# smart function inspired by py-SMART https://github.com/freenas/py-SMART
+# under Copyright (C) 2015 Marc Herndon and GPL2
+
+"""Displays HDD smart status of different drives or all drives
+
+Parameters:
+    * smartstatus.display: how to display (defaults to "combined", other choices: "seperate" or "singles")
+    * smartstauts.drives: in the case of singles which drives to display, separated comma list value, multiple accepted (defaults to "sda", example:"sda,sdc")
+"""
+
+import os
+
+import bumblebee.util
+import bumblebee.output
+import bumblebee.engine
+
+from shutil import which
+from subprocess import Popen, PIPE
+
+
+class Module(bumblebee.engine.Module):
+    def __init__(self, engine, config):
+        super(Module, self).__init__(engine, config, None)
+        self.devices = self.list_devices()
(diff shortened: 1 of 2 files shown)


martindoublem:
I could potentially add a feature to add new drives after the launch too but I dont have the need for it myself.

bumblebee/modules/smartstatus.py
```diff
@@ -0,0 +1,117 @@
...
+            if result == 'Fail':
+                return 'Fail'
+            if result == 'Pre-fail':
+                return 'Pre-fail'
+        return 'OK'
+
+    def list_devices(self):
+        for (root, folders, files) in os.walk('/dev'):
+            if root == '/dev':
+                devices = {"".join(filter(lambda i: i.isdigit() == False, file)) for file in files if 'sd' in file}
```
tobi-wan-kenobi: For a future extension, it might be nice to support also other naming schemes than "sdX" here? I have an NVMe, for instance, so my root disk is called /dev/nvme0n1

tobi-wan-kenobi:
That's a really cool module, thanks for adding!

bumblebee/modules/smartstatus.py
martindoublem: As I mentionned in the first line, the smart function was inpired from py-smart with that line specifically being more or less a copy with the exception of sudo that I added, I did see that in there and thought about it and remembered that was a disk format, it completely missed my attention when I was created the list devices as I only have sdX devices, is there any other devices I should be paying attention to?

martindoublem: Could you let me know what you have to enter to get the smart status of your drive?
 `sudo smartctl -a /dev/XXXX`

I mean it wont take much for me to do, I just need to know the path of the devices to watch out for, as I currently do not have one. I have no idea.

tobi-wan-kenobi: @martindoublem I use `sudo smartctl -a /dev/nvme0n1`

PR #591, opened 2020-04-05
Fixing no internet crash (Sun Module)

soykan:
error
Using sun module without internet crashes the status bar. With this patch instead of crashing it will show a question mark.
second


tobi-wan-kenobi:
Looks good to me, thanks for fixing!

PR #596, opened 2020-04-10
[smartstatus] Added support for nvme SSDs

martindoublem:


tobi-wan-kenobi:
Looking good, thanks a lot for the PR!

PR #598, opened 2020-04-16
Clickable Todo Module

soykan:
With this patch user can show and edit todo list by left mouse click. It will open the default text editor with xdg-open.


tobi-wan-kenobi:
Nice addition, many thanks!

Issue #612, opened 2020-05-01
[dev] no icons showing for sensors2


martindoublem:
### Bug Report
Icons are not showing. I have spent some time trying to fix it, but I can't seem to figure out how it is handled.

#### Summary
Affected module: sensors

#### Description:
<description>

#### How to reproduce
bumblebbe-status -m sensors2 -t iceberg

martindoublem:
Can you check if it is an issue with icons on modules with multiple widget. I can seem to get them to work in smartstatus after adding icons to the json as well.

martindoublem:
I have been investigating, sadly going nowhere.

tobi-wan-kenobi:
This is a stupid bug from my side, widgets *need* a reference to their module.

I'll fix it in sensors for now, but probably it makes sense to add a "generator" method to the module that produces widgets, so that this issue doesn't pop up more often.

Thanks for reporting!

martindoublem:
well I have the same in smartstatus once I add an icon.

PR #615, opened 2020-05-01
[modules/smartstatus] Add module reference to widgets

martindoublem:
fixes related to #612


tobi-wan-kenobi:
Thanks a lot! :+1: 

I've also added a `add_widget()` helper function to the `Module` class that should help in these situations in the future.

Issue #617, opened 2020-05-02
[dev] modules.contrib.nvidiagpu is broken when there's a Nvidia GPU


somospocos:
#### Summary
Affected module: contrib.nvidiagpu

#### Description:

Happens in revision a8ab4f95 for me.

    File "/mnt/d2/downloads/work/bumblebee-status/modules/contrib/nvidiagpu.py", line 28, in hidden
    if "not found" in self._utilization.startswith("not found"):
    AttributeError: 'Module' object has no attribute '_utilization'

First glance shows a naming inconsistency most likely due to a typo, because the name lacks an underscore compared to the one from `Module.__init__()`

Adding an underscore produces a different error:

    File "/mnt/d2/downloads/work/bumblebee-status/modules/contrib/nvidiagpu.py", line 28, in hidden
    if "not found" in self.__utilization.startswith("not found"):
TypeError: argument of type 'bool' is not iterable

A deeper look at this of course shows there's something wrong with it.

I'll try to submit a pull request that fixes it for me in a few minutes, but I have no way to test it on a setup when the Nvidia GPU is missing, try testing that use case too if possible before merging, it would be a shame if the change that fixes it for me will break it for others.

tobi-wan-kenobi:
I don't have an Nvidia GPU, I'll just test the PR before merging :)

tobi-wan-kenobi:
Awww, I see the issue.

Don't bother with a PR, that's a stupid one

somospocos:
What an unfortunate timing, I already made the pull request :)

My change is slightly different though, I check for the presence of not found string at the end, that's how I understood the code.

I suppose I can just delete my fix branch then?

tobi-wan-kenobi:
Yes, please. I was really torn between the two solutions (as the initial bug clearly demonstrated), but I think searching for 'not found' in the whole string *should* make the solution slightly more stable when the tool's output changes slightly.

Thanks for the quick PR, though!

somospocos:
Updated the repo after your fix and bumblebee-status runs correctly, feel free to close.

PR #618, opened 2020-05-02
[modules/nvidiagpu] fix module

somospocos:
Fixes https://github.com/tobi-wan-kenobi/bumblebee-status/issues/617


tobi-wan-kenobi:
:) I fixed it using `'not found' in self.__utilization` - would that also be fine for you?

PR #620, opened 2020-05-02
Octoprint

bbernhard:
Not sure if that's useful for anybody, but as it's working quite well so far, I thought I could add it :)

Selection_066

### DIFF
--- a/modules/contrib/octoprint.py
+++ b/modules/contrib/octoprint.py
@@ -57,7 +57,7 @@ class WebcamImagesWorker(threading.Thread):
         self.__running = False
 
 class Module(core.module.Module):
-    @core.decorators.every(minutes=1)
+    @core.decorators.every(seconds=5)
     def __init__(self, config, theme):
         super().__init__(config, theme, core.widget.Widget(self.octoprint_status))
 
--- a/modules/contrib/octoprint.py
+++ b/modules/contrib/octoprint.py
@@ -1,0 +1,224 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the Octorpint status and the printer's bed/tools temperature in the status bar.
+
+   Left click opens a popup which shows the bed & tools temperatures and additionally a livestream of the webcam (if enabled).
+
+Parameters:
+    * octoprint.address     : Octoprint address (e.q: http://192.168.1.3)
+    * octoprint.apitoken    : Octorpint API Token (can be obtained from the Octoprint Webinterface)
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thanks a lot!

Hope you liked the refreshed APIs :)

I'll merge the PR right away, but I have a couple of optional suggestions/questions:

- You don't need to call `update_status()` in the constructor of the module, that will be done automatically during the first update
- Is there a specific reason for not renaming `update_status()` to `update()` and remove one level of indirection?
- Very minor: I tried to use single-quoted strings throughout the whole project for consistency. But TBH, I'm thinking about just using a source code formatter like black or yapf...

bbernhard:
> Thanks a lot!
> 
> Hope you liked the refreshed APIs :)

yep, very nice. really like the `every` decorator :+1: 

> 
> I'll merge the PR right away, but I have a couple of optional suggestions/questions:
> 
>     * You don't need to call `update_status()` in the constructor of the module, that will be done automatically during the first update
> 
>     * Is there a specific reason for not renaming `update_status()` to `update()` and remove one level of indirection?

good catch - I'll create a PR for that, thanks! :)

>     * Very minor: I tried to use single-quoted strings throughout the whole project for consistency. But TBH, I'm thinking about just using a source code formatter like black or yapf...

If the single quote is preferred, I can easily change that. I do not have a strong preference for one or the other, so just let me know. But I agree with you, a common coding style among all the plugins would be great.

As you mentioned `black`: I have contributed a bit to the HomeAssistant project over the past few months and they use `black`, `isort`, `tox`, etc. together with some pretty strict coding style guidelines in their project. While it is sometimes a bit frustrating to follow all the rules (e.g every method needs to have a doc string, every new module needs to have unittests, python imports need to be properly ordered, etc), it definitely helps to get a consistent style across all plugins (and I think that definitely pays off in the long run :))

**edit:** I _think_ `black` only supports double quotes - afaik they dropped single quote support in favor of double quotes.

tobi-wan-kenobi:
Didn't know `isort` and `tox`, will look into those, thanks for the hint!

I've now adopted the `black` style throughout the project, so nobody should have to care about formatting anymore :)

Regarding guidelines: I want to be really relaxed about modules, those are essentially free-format to me, because I want to make contributing as easy as possible (and I know how frustrating it is to contribute your spare time to a project that throws back your changes with "please follow whatever style I think is best").
For core changes, I'll be more ... demanding, probably :)

And you know my take on stuff like doc strings: If you have to *document* what your method does, it probably deserves a better name first :P

PR #625, opened 2020-05-07
Improved three modules in showing more compacted info with new parameters

martindoublem:


modules/contrib/smartstatus.py
```diff
@@ -29,6 +30,7 @@ def __init__(self, config, theme):
         self.devices = self.list_devices()
         self.display = self.parameter("display", "combined")
         self.drives = self.parameter("drives", "sda")
+        self.show_names = self.parameter("show_names", "True")
```
tobi-wan-kenobi, line 33: For this, you can use `util.format.asbool()` to unify boolean handling - that automatically allows a user to use yes/no, true/false, 0/1 for the parameter.

modules/core/redshift.py
```diff
@@ -74,6 +75,7 @@ def __init__(self, config, theme):
         super().__init__(config, theme, core.widget.Widget(self.text))
 
         self.__thread = None
+        self.transition_info = self.parameter("transition_info", "True")
```
tobi-wan-kenobi, line 78: same here (`util.format.asbool()`)

modules/contrib/smartstatus.py
martindoublem: Man, I was wondering why the True False boolean was not working in your code!

martindoublem:
Actually @tobi-wan-kenobi can you finish the last few changes, I am not sure how to use that util.format.asbool() in the conditional statements.

tobi-wan-kenobi:
Changes pushed.

I also took the liberty of renaming "transition_info" in redshift to "show_transition", I think that captures the intent a bit clearer. I hope that is fine with you - if you have objects, please let me know, and I will gladly change the code again.

Thanks for the contribution!

martindoublem:
You are always welcome, you know how much I like to use your status bar!

PR #630, opened 2020-05-15
Handle n+1 characters in the scrolling decorator

TheEdgeOfRage:
Hopefully the last fix for cmus :smile: 

When the text is 1 character longer than the width of the module it scrolls once to the right, once to the left and then to -1 ending up with an empty string in the output. Then returns back to 0 and starts over again.

I wrote a test, found the edge case, and hopefully fixed it with this.

PS. I also added `.mypy_cache` to the gitignore file. It's a static type checking tool that keeps its cache in the repo. I recommend it for type annotated code checking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ ENV/
 
 # Visual studio project files
 .vscode/
+
+# mypy cache
+.mypy_cache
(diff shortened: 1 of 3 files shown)


tobi-wan-kenobi:
Looks good to me, many thanks for the improvement - and particularly for adding tests for it!

PR #631, opened 2020-05-17
Fix network_traffic mobule on development

zerorust:
When I shifted over to the development branch after reading some of the recent activity on github, I found a problem with the `network_traffic` module.  Here is a simple PR that fixes this immediately problem but there is a bigger issue with `development:core.widget.Widget`.  Namely, if a module uses the `module=` arg in the `__init__` then it is going to crash because there is some sort of a circular reference in the `index()` call needed for setting the `widgets` parameter. 

I have only look at your code for a half-hour, so I do cannot talk in more detail.  I'd be happy to look at it more if needed.  It seems that no other modules use the `module=` arg to `__init__`...

Thanks for a great tool.

### DIFF
--- a/bumblebee_status/modules/contrib/network_traffic.py
+++ b/bumblebee_status/modules/contrib/network_traffic.py
@@ -17,6 +17,7 @@ import util.format
 
 WIDGET_NAME = "network_traffic"
 
+
 class Module(core.module.Module):
     def __init__(self, config, theme):
         widgets = [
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Ah, I see the issue. Yes, thanks for raising this!

I think the "module" parameter in the widget is obsolete now, that was a bit of a hack to begin with.

tobi-wan-kenobi:
OK, I have to qualify that: Since some modules (xrandr, eg.) break when I remove that parameter, it obviously *is* still used. I'll take that up for the next round of refactoring, for sure.

zerorust:
Ok, great.

PR #635, opened 2020-05-26
[modules] New amixer module with input support

ardadem:
This is basically amixer module with input support. I renamed it as amixer2 because some people can just want use lightweight amixer module.


tobi-wan-kenobi:
Thanks a lot for the addition!

The changes look totally fine to me, I will actually just move over that module over the existing amixer, as it's a great improvement.

If you object, please let me know and I will revert those changes.

PR #637, opened 2020-05-28
[modules] GitHub unread notifications module with modular reason support

cristianmiranda:
Hi! :wave: First, I want to thank all who are working on this amazing bar. I enjoy it very much and module writing is very easy! :heart: 

I did this module because I needed it in order to quickly identify the source of unread notifications since they don't have all the same urgency to me.
Here are some screenshots of how it's configured, how it looks like and how they match GitHub notifications page. I created a few other modules that I also needed but wanted to create my first PR with this one just to see how it goes.

Notifications' reasons docs: https://developer.github.com/v3/activity/notifications/#notification-reasons

---

Updated screenshot after going with this suggestion: https://github.com/tobi-wan-kenobi/bumblebee-status/pull/637#issuecomment-635511914

image

---


tobi-wan-kenobi:
Very cool, thanks a lot for the contribution!

One idea: If you made it so that the module would just sum up all notifications when no reasons are specified, this module could *replace* the current `github` module and stay backwards-compatible.

I think that would be nice, so that people don't have to wonder which of the two github modules to use.

What do you think about that?

cristianmiranda:
Excellent idea. I'll update the PR with your suggestion. Should I replace the current github module (including icon fonts) or is that something you want to handle by yourself?

tobi-wan-kenobi:
Feel free to just replace it :+1: 

Please just leave in the original author's contribution (yvesh) and add your own, for fairness' sake.

Thanks!

cristianmiranda:
@tobi-wan-kenobi what do you think about keeping original github module's count as it is and add "reason" counts next to it only if configured? Something like the following:

* No reasons configured: `8`
* Few reasons configured: `8 - 1/1/3`

I'm bringing this up because now that you mention the old module, I don't thing it'd be a good idea losing the total unread count (even though users configure reasons to be displayed).

tobi-wan-kenobi:
I think that's an excellent idea!

cristianmiranda:
> I think that's an excellent idea!

Excellent! I'll update the PR once again. Thanks!

tobi-wan-kenobi:
Also, since I noticed it (small stuff):
- For splitting the string, you can use `util.format.aslist` (that way, all lists have the same format, always)
- If you don't mind, I'll run the module through `black` afterwards to ensure uniform style (e.g. double quotes instead of single)

cristianmiranda:
> Also, since I noticed it (small stuff):
> 
>     * For splitting the string, you can use `util.format.aslist` (that way, all lists have the same format, always)
> 
>     * If you don't mind, I'll run the module through `black` afterwards to ensure uniform style (e.g. double quotes instead of single)

Perfect, I'll use that and, of course! I'm not familiar with `black` but if you point me in the right direction I could also run it myself before submitting new code. Actually, that'd be great since I'm planning on contributing some other modules in the future :smile:

tobi-wan-kenobi:
Great, I am looking forward to your contributions!

Please bear with me if reviews sometimes take a while, times are pretty busy right now and this is strictly a spare time project (though one very important to me).

Reg. black: https://github.com/psf/black

Basically, it's just `black -t py34 <filename>` and you should be good :-)


Thanks a lot!
⁣Sent from TypeApp ​

cristianmiranda:
I totally understand. Thank you, again! :heart: 

I ran the module through `black` and submitted my changes. I'll update the PR screenshots so you can use them for documentation.

image

tobi-wan-kenobi:
Very cool, thanks a lot!

I will merge it now, I'm very happy with how this module turned out, will definitely use it myself :)

One thing to consider: With other modules (APT, Arch, etc.) people were/are complaining that the compact format a/b/c/d makes it hard to figure out what's going on. So maybe having an "extended" output format by default (e.g. PR: 5 Mention: 1) would make sense - and just not show anything that's 0?

tobi-wan-kenobi:
Also: "development" has been merged to master recently, so future merges can go into master. I will delete development now, TBH, I just forgot :)

fredj:
very nice improvement ! thanks @cristianmiranda

PR #646, opened 2020-06-03
[module] playerctl to displays information about the current song in …

smitajit:
…vlc, audacious, bmp, xmms2, spotify and others


tobi-wan-kenobi:

Unfortunately, the module itself uses old APIs (e.g. the engine, and the whole "bumblebee." prefix).

Migrating to the new APIs shouldn't be a lot of effort, since it's mostly renamings, but right now, the module wouldn't work properly with the "master" version of bumblebee-status.

Having said that, I'd *love* to have this module in, it's a great idea!

tobi-wan-kenobi:
Awesome, thanks a lot!

Goes straight into master :)

PR #649, opened 2020-06-07
Monitor can be set from cpu module

P-Storm:
I'm using an different system monitor, and couldn't set this up. So I added a command line option. I'm also checking if the user input is a valid proces to be executed. :)


tobi-wan-kenobi:
Thanks for the PR!

Actually, you should be able to override this behaviour for any module via CLI parameters already:

`bumblebee-status -m cpu -p cpu.left-click=<whatever>`

Basically, the modules just define "default bindings" that the user can *always* override. So, if that is fine with you, I'd close the PR.

Please let me know in case this doesn't work, because that would be a bug then :)

P-Storm:
Yes this may be closed, didn't read the manual good enough. For other people who search this issue, the info can be found:: https://bumblebee-status.readthedocs.io/en/master/features.html#events

tobi-wan-kenobi:
@P-Storm Thanks for raising the point, though, this made me discover #650  :)

PR #658, opened 2020-06-21
fix a regression for shell subprocess

es80:
Hello,

I was using the `shell` module to run a command which used a subshell. 

Unfortunately after the last update to `bumblebee-status` this stopped working. The `$(...)` in my command was interpreted literally instead of spawning a subshell. 

Putting `shell=True` in the call to `util.cli.execute` in `shell.py` fixed the problem for me. 

I see on SO that there may be security concerns related to shell injection when using `shell=True` so perhaps there is a better way to fix my problem? I'm not too sure.

Thanks,
Elliott.


tobi-wan-kenobi:
Good catch, thanks! I'll release 2.0.1 for that :)

PR #660, opened 2020-06-22
Do not execute iwgetid if the interface is recognized as tunnel

### LINKED ISSUE #659: (nic) Tunnel causes problems with iwgetid
#### Description
Affected module: nic
Version used: latest AUR (2.x)

After I've attempted to update to the latest Bumblebee release there seems to be an issue with the `nic` module. I'm connected to a WireGuard VPN using the network-manager and while the interface gets correctly recognized as tunnel and the device name as well as the IP is shown, the module states that `/usr/bin/iwgetid -r wg0 exited with code 255`. 

I guess this command shouldn't be executed when dealing with a tunnel, should it?

#### How to reproduce
- Install Bumblebee from AUR (latest 2.x version)
- Use `nic` module
- Setup or import WireGuard tunnel using network-manager
- Enable connection e.g. by `nmcli con up wg0`

thannaske:
Fixes #659

`iwgetid` is used to gather the SSID of a wireless network interface. However due to the fact that some interfaces may start with a `w` that no wireless interfaces, we should execute `iwgetid` only if the interface is not recognized as tunnel as well.

### DIFF
--- a/bumblebee_status/modules/core/nic.py
+++ b/bumblebee_status/modules/core/nic.py
@@ -125,7 +125,7 @@ class Module(core.module.Module):
             widget.set("state", state)
 
     def get_ssid(self, intf):
-        if self._iswlan(intf) and self.iwgetid:
+        if self._iswlan(intf) and not self._istunnel(intf) and self.iwgetid:
             return util.cli.execute(
                 "{} -r {}".format(self.iwgetid, intf), ignore_errors=True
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks a lot for the contribution, that looks spot-on!

I'll push the update to AUR and PyPI as well.

Issue #577, opened 2020-03-20
Spotify buttons


richibrics:
### Feature Request
Hi, I wonder if it is possible to add play/pause, next and previous buttons to the spotify module.

martindoublem:
Already implemented, did you try left click, right click, middle click on it?

richibrics:
Yes, and they work perfectly, but I was thinking about implementing buttons that you can see in this bar: ![image](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/screenshots/themes/iceberg-dark-powerline.png)

martindoublem:
Ha, I see. If I am not wrong that is another player but I am not sure which.

perhaps a parameters could be added to make these extras buttons visible as I quite like it without it (I have so many modules running, it is taking almost my entire screen width.)

Maybe you can look into it, if you have the knowledge in Python?

martindoublem:
I just took a look at the module and it looks like a whole new module would be needed so probably not easily doable.

richibrics:
Thanks anyway

martindoublem:
You can leave your request open, it is something, I could do eventually, but I won't lie as it is not something I need, it may wait a bit before I get to it.

tobi-wan-kenobi:
I like the idea a lot! However, since I am not using spotify myself, I'll have to wait for a PR from somebody else, which I will be happy to review and merge :)

richibrics:
> Ha, I see. If I am not wrong that is another player but I am not sure which.
> 
> 
> 
> perhaps a parameters could be added to make these extras buttons visible as I quite like it without it (I have so many modules running, it is taking almost my entire screen width.)
> 
> 
> 
> Maybe you can look into it, if you have the knowledge in Python?

Maybe I will take a look at it

martindoublem:
@tobi-wan-kenobi do you know which module has those buttons?

tobi-wan-kenobi:
@martindoublem cmus has them for sure, and mpd as well, I think. mpd even has a fancy "layout" parameter that allows to specify which widgets to show.

somospocos:
Both cmus and mpd modules have a layout parameter. IIRC cmus had that feature originally, and I liked it a lot, so I refactored the mpd module to add that as well, because being able to show a subset of widgets is a great way to save space in the bar + being able to specify their order makes the visual look of the module more customizable. Unfortunately I don't use spotify at all, so I can't help with this :(

[...]

PR #663, opened 2020-06-24
Spotify buttons module

LtPeriwinkle:
I saw issue #577 and thought it seemed like an interesting idea, so I copied the original Spotify module and added widgets for playback control to it. I'm not exactly a python genius, so it probably isn't as good as someone who is would do, but I thought it might be useful.


tobi-wan-kenobi:
Very, very cool, thanks a lot!

2 minor suggestions/requests:

- For the layout, could you please use `util.format.aslist()`? That way, all lists are uniform (comma-separated) - that's particularly because argument parsing otherwise requires you to put the layout in quotes, I believe.
- I know it's tempting to split this out into a separate module, but to avoid confusiong for users (and I think buttons are useful for *everyone*, would you be OK if this simply replaces the current spotify module?

Thanks a lot for your contribution!

martindoublem:
>     * I know it's tempting to split this out into a separate module, but to avoid confusiong for users (and I think buttons are useful for _everyone_, would you be OK if this simply replaces the current spotify module?

Ideally having the buttons as a parameter and the text as another would be better if you ask me, with both enabled by default would be great!

tobi-wan-kenobi:
Looks good to me, thanks! In the future, it *might* be worth to remove the hardcoded unicode characters and replace them with theme info, like `cmus` does, for example, but other than that, this looks great. Many thanks!

LtPeriwinkle:
Something a little strange (this might even be isolated to my machine), but since I overwrote the original spotify with the new version, most clicks on a button seem to register many times (for example skipping the entire playlist instead of one song, or playing and pausing super fast until the music just sounds like static) and it even blocks playing and pausing on the spotify window itself. I don't know what would cause this though (or how to fix it), and in case it's not isolated, I just wanted to let you know.

tobi-wan-kenobi:
OK, that's good to know, thanks. I'll look into the code whether I spot something.

PR #667, opened 2020-06-28
[modules/spotify] Icon improvements

gkeep:
1. Added icons for different states to awesome-fonts and ASCII icon sets
2. **Improved icon handling**:
    1. They were hardcoded before, which is terrible for a lot of reasons
    2. Because of previous point, only ASCII icon set was supported

image

By the way, you should document this stuff, I wasted a couple of hours trying to find info about this. While docs mention widget states, they don't mention anything about icons in there.


tobi-wan-kenobi:
Thanks for the contribution, great improvement! I agree that hardcoded icons are not a good idea, I think that was a first draft to get the module going, also due to the fact as - like you've noted - the documentation leaves to be desired wrt. icon handling.

> By the way, you should document this stuff, I wasted a couple of hours trying to find info about this. While docs mention widget states, they don't mention anything about icons in there.

Duly noted, I will track that as #669 - if the explanation is inaccurate, please let me know.

PR #671, opened 2020-06-29
Fix indentation error leading to missing output

rassie:


rassie:
Ah, it's already on `main`. Sorry for the noise.

PR #672, opened 2020-06-29
[modules/spotify] Fix layout parameter

gkeep:
While working on icons yesterday, I noticed that `spotify.layout` parameter did not work. This PR fixes it.


tobi-wan-kenobi:
Thanks for the change and noticing the problem!

Damn, that slipped in during an earlier change request I had for a PR, and I failed to see that during the review.

I'd still prefer to use `util.format.aslist()`, I'll just update the code directly.

tobi-wan-kenobi:
fixed in f75f1a9

PR #673, opened 2020-06-30
Convert spotify.pause to an action

rassie:
This is probably controversial and bike-shedding at its best, so feel free to ignore.

In the recent weeks, widget buttons have replaced mouse actions in the `spotify` module. However, I feel the logic for the play/pause button is confusing. The buttons `next`, `prev` are currently actions, however the button `play`/`pause` is reflecting a state, i.e. when playing a `playìng` triangle is displayed and also when paused a `paused` double-line. If implemented as an action instead like the other buttons ("please pause", "please play"), the widget icons would be the exact opposite: a triangle when paused and double-line when playing and indeed this is exactly the pattern every other player (and especially Spotify itself) implements.

This PR switches the icons to reflect a desired action instead of state.

### DIFF
--- a/bumblebee_status/modules/contrib/spotify.py
+++ b/bumblebee_status/modules/contrib/spotify.py
@@ -89,9 +89,9 @@ class Module(core.module.Module):
                         ).Get("org.mpris.MediaPlayer2.Player", "PlaybackStatus")
                     )
                     if playback_status == "Playing":
-                        widget.set("state", "playing")
-                    else:
                         widget.set("state", "paused")
+                    else:
+                        widget.set("state", "playing")
                 elif widget_name == "spotify.next":
                     widget_map[widget] = {
                         "button": core.input.LEFT_MOUSE,


gkeep:
I agree with you. That's an aspect I totally overlooked while doing #667.

I was inspired by other media modules (playerctl, mpd) while making said PR, which is why spotify module currently displays the state.

rassie:
You are probably right, however, I've got no experience with icon sets yet, will have to dig in a bit. Do you know your way around them? If you are more fluent, maybe you could craft up a PR, then I'd close this one.

gkeep:
Here they are:
* ASCII:
https://github.com/tobi-wan-kenobi/bumblebee-status/blob/f75f1a9f26c6a4b96251f076cfd369e7fc276cab/themes/icons/ascii.json#L302
https://github.com/tobi-wan-kenobi/bumblebee-status/blob/f75f1a9f26c6a4b96251f076cfd369e7fc276cab/themes/icons/ascii.json#L303
* Awesome-icons:
https://github.com/tobi-wan-kenobi/bumblebee-status/blob/f75f1a9f26c6a4b96251f076cfd369e7fc276cab/themes/icons/awesome-fonts.json#L224
https://github.com/tobi-wan-kenobi/bumblebee-status/blob/f75f1a9f26c6a4b96251f076cfd369e7fc276cab/themes/icons/awesome-fonts.json#L225

Not sure if you can attach them to this PR in web-editor though.

tobi-wan-kenobi:
Great addition, thanks! I agree it makes sense that the buttons model actions (for all media modules, actually)

I also agree with @gkeep 's assessment that changing the icon set files is the best approach.

Thanks a lot!

gkeep:
I can cook up a PR that changes states to actions in all media modules, if everyone is okay with that?

rassie:
To be continued in #674, closing here.

PR #675, opened 2020-06-30
[modules/nic] Using `iw` to find out whats the SSID name

xsteadfastx:
Using iw and not the deprecated `iwgetid`.


tobi-wan-kenobi:
Oh, that is a good change, many thanks!

PR #677, opened 2020-07-05
Small fix to xrandr module.

andrewreisner:
This fixes a few issues I had using the xrandr module (using python 3.6 and 3.7).


tobi-wan-kenobi:
Oh, those are really important fixes :) Thanks a lot for the PR!

Issue #678, opened 2020-07-06
Read theme & modules from the config file


fenilgandhi:
### Feature Request

Currently the config file only supports configuring individual moduiles for bumblebee-status
But i wanted to also specify my theme & list of modiules , right into the config file. I tried searching for a way to do it but couldn't find any. Having these values in a default config file, saves the effort of maintaining consistency if the user invokes the "bumblebee-status" from different application scripts.

fenilgandhi:
@tobi-wan-kenobi : I could not find any developer documentation for contributing to the project itself. Can you atleast specify  the minimal things how to generate a build

tobi-wan-kenobi:
That's a nice FRQ, thanks!

OK, so that is true, I didn't document that. You don't need a build, just check out the git and run bumblebee-status :)

adgai19:
> ### Feature Request
> Currently the config file only supports configuring individual moduiles for bumblebee-status
> But i wanted to also specify my theme & list of modiules , right into the config file. I tried searching for a way to do it but couldn't find any. Having these values in a default config file, saves the effort of maintaining consistency if the user invokes the "bumblebee-status" from different application scripts.

Can someone add some documentation for this?? I am really excited for this feature.

tobi-wan-kenobi:
should be in the "Features" section now

PR #679, opened 2020-07-06
Updating config parser

fenilgandhi:
Towards #678 

I am not sure if this is the way to implement this feature. But i would be happy to receive some guidance on it.


tobi-wan-kenobi:
That looks good to me, thanks for the contribution!

2 things:

* Before merging the PR, I want to do some preliminary testing, so it'll be around the weekend until I merge - apologies for the delay
* It might be more "principle of least surprise" to load the modules from the config and then add whatever is specified on the commandline - what do you think?

PR #680, opened 2020-07-10
Messagereceiver

bbernhard:
Displays data received via unix sockets. 

I am quite often creating gif screencasts. And the one thing that was always missing was some sort of visual indication whether a recording is running or not. With this integration I can now send data from my gif recoding bash script to bumblebee to let it know when a recording started/stopped :) 

Hope it's also useful to others.

### DIFF
--- a/bumblebee_status/modules/contrib/messagereceiver.py
+++ b/bumblebee_status/modules/contrib/messagereceiver.py
@@ -1,0 +1,85 @@
+# pylint: disable=C0111,R0903
+
+"""
+Displays the message that's received via unix socket.
+
+Parameteres:
+    * messagereceiver   : Unix socket address (e.g: /tmp/bumblebee_messagereceiver.sock)
+"""
+
+import core.module
+import core.widget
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Nice and generic, very cool addition, thanks!

Just a short heads-up: With the new bumblebee-code, I think there's actually an even more elegant way to accomplish what you are doing:

If you set `self.background` to `true` in `__init__()`, this will *automatically* call `update()` in a separate thread and wait for it to finish. That way, you can just listen for incoming messages in `update()` (without any need of a queue or something like that) and whenever you *do* get a message (i.e. after processing in `update()` is finished), the surrounding scaffolding will automatically call `core.event.trigger("update", [module.id], redraw_only=True)` to enforce a refresh of the module.

tobi-wan-kenobi:
Oh, and just to clarify: The whole advantage of the above is that you can then set the module's default update interval to `never` and it will always update "just in time", which reduces CPU consumption.

PR #681, opened 2020-07-11
reworked messagereceiver module

bbernhard:
Many thanks for the tip with `self.background`, that really reduced the LOC significantly. Great to see that such a nice thing is already built in the core - that makes writing plugins even easier :) 

I hope I am using it correctly. If not, please let me know.

### DIFF
--- a/bumblebee_status/modules/contrib/messagereceiver.py
+++ b/bumblebee_status/modules/contrib/messagereceiver.py
@@ -18,29 +18,36 @@ Example:
 
 """
 
+import socket
+import logging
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks :) Yeah, I noticed this "I have a long-running task, so I spawn a separate thread and make it communicate with the module" was becoming a pattern, so I thought it might be useful general-purpose (see arch-update or ping, for example).

Yes, you are using it correctly, I'll go ahead and merge :)

PR #682, opened 2020-07-12
Add simple Gentoo portage status module.

andrewreisner:
This adds a module to display the status of Gentoo portage operations by reading the portage logfile.


tobi-wan-kenobi:
Cool addition, many thanks!

Just a suggestion: It might be worth to set a higher default interval (1-5m?) for this module, so that the check isn't performed every second.

PR #683, opened 2020-07-18
modules.rst: Add more module dependency info

jayvdb:


jayvdb:
Thanks for the great work on v2, especially documentation of the dependencies, without which these minor improvements would have been impossible.

I've been able to get most of the dependencies into openSUSE , and now https://build.opensuse.org/request/show/821557 for correctly mapping modules to their varied dependencies, and get the test suite operational.

tobi-wan-kenobi:
Thanks a lot for the contribution, and for the openSUSE build! Was wondering who did those :)

Just one note: I've neglected to document that anywhere (and I will presently fix that), but "modules.rst" is actually auto-generated from the modules' docstrings, so we need to put your comments into the corresponding modules.

`bumblebee-status -l modules-rst > docs/modules.rst` then produces the final documentation.

tobi-wan-kenobi:
@jayvdb added the changes you proposed (with minor modifications) to the docstrings, hope that is fine with you!

Thanks for your efforts!

jayvdb:
Ya of course that is fine.  I should have noticed that it was generated - I did get the feeling that there was commonality between the docstrings and the rst.

> bumblebee-status -l modules-rst > docs/modules.rst

Perhaps put that line into CI, followed by `git diff --exit-code` so it fails if someone edits it directly, and they can see the cause.

PR #684, opened 2020-07-25
[xrandr] unit tests, exclude parameter, and final display safeguard

nayaverdier:
Also a minor change in `util/popup.py` to only add the `Close` button if `leave=False`

### DIFF
--- a/bumblebee_status/modules/core/xrandr.py
+++ b/bumblebee_status/modules/core/xrandr.py
@@ -31,7 +31,7 @@ import util.format
 
 try:
     import i3
-except ModuleNotFoundError:
+except Exception:
     pass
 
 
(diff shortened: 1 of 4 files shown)


bumblebee_status/modules/core/xrandr.py
```diff
@@ -40,40 +40,50 @@ class Module(core.module.Module):
     def __init__(self, config, theme):
         super().__init__(config, theme, [])
 
+        self._exclude = tuple(filter(len, self.parameter("exclude", "").split(",")))
```
tobi-wan-kenobi, line 43: instead of manually splitting here, you could use the method util.format.aslist(), which gives uniform list formatting (incidentially, as comma-separated list)

tobi-wan-kenobi:
Very cool changes, many thanks for your contributions!

The diff looks really good to me, as I am a heavy xrandr user myself, I will go ahead and commit the PR, and test it carefully during my daily usage :-)

nayaverdier:
Awesome, thank you! Anything you'd like added to the tests?

tobi-wan-kenobi:
Hm, the only thing I can think of is testing that during runtime, disconnected/connected displays are picked up correctly, but I think that doesn't have a very high priority.

Thanks for adding those tests that you added already, that goes a long way improving a somewhat fragile module :-)

PR #686, opened 2020-07-26
[xrandr] Add autotoggle behavior to xrandr

nayaverdier:
More xrandr changes! I removed `_needs_update`, so i3 subscription events and widget clicks will update immediately. Not sure if there was a reason for having it delayed, but seems to work well on my machine. Let me know if there's a reason to not update right away, I'll add that back.

### DIFF
--- a/bumblebee_status/modules/core/xrandr.py
+++ b/bumblebee_status/modules/core/xrandr.py
@@ -9,6 +9,8 @@ Parameters:
     * xrandr.autoupdate: If set to 'false', does *not* invoke xrandr automatically. Instead, the
       module will only refresh when displays are enabled or disabled (defaults to true)
     * xrandr.exclude: Comma-separated list of display name prefixes to exclude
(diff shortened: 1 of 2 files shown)


bumblebee_status/modules/core/xrandr.py
```diff
@@ -35,91 +37,147 @@
...
+            core.input.register(widget, button=3, cmd=self._toggle)
+
+            widget.set("state", display.state)
+            widget.set("pos", display.position)
 
         if not self._autoupdate:
             widget = self.add_widget(full_text="")
             widget.set("state", "refresh")
-            core.input.register(widget, button=1, cmd=self._refresh)
+            core.input.register(widget, button=1, cmd=self.update)
```
tobi-wan-kenobi, line 145: you could use core.input.LEFT_MOUSE instead of button 1 here, for clarity

tobi-wan-kenobi:
I think the needs_update was a left-over from an earlier version. Thanks a lot for your changes, they look really good!

PR #687, opened 2020-07-31
[modules/contrib/stock] handle urllib request exception

anlif:
Handle exception that is raised when e.g. your network connection is
down.


tobi-wan-kenobi:
Looks good, thanks a lot for the contribution!

PR #688, opened 2020-08-06
Fix wrong usage of asbool in layout-xkb

alexsr:
In Line 56 the `util.format.asbool` method was used with 2 instead of 1 parameter due to a misplaced `)`.
This caused the module to fail and therefore the json passed to i3 was malformed.


tobi-wan-kenobi:
Oh, good catch, thanks for the contribution!

alexsr:
Thank you for your work on this cool project.
Easy to use, great functionality and nice to look at!

PR #689, opened 2020-08-10
Force status bar to update early using USR1 signal

joshbarrass:
This is something I added a [few years ago](https://github.com/tobi-wan-kenobi/bumblebee-status/pull/255) that seems to have been lost over the years. Everything seems to have changed a lot since then, but this PR reintroduces this functionality as it is something that has always been very useful to me and may continue to be useful to others.

I've no idea whether this functionality was removed intentionally; if that is the case, I understand rejecting this PR. 

If there is anything that needs tweaking (perhaps I've committed some unthinkable crime against bumblebee), point me in the right direction and I will fix it as soon as I can.

### DIFF
--- a/bumblebee-status
+++ b/bumblebee-status
@@ -4,6 +4,7 @@ import os
 import sys
 import json
 import time
+import signal
 import socket
 import select
 import logging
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Oh, sorry! That was unintentionally removed, my apologies!

Thanks for the PR, looks good!

PR #690, opened 2020-08-10
Quotes around the deadbeef string to prevent parsing errors by deadbeef

joshbarrass:
If using complex queries, deadbeef can fail to parse the query properly if it is not surrounded by quotes. This PR surrounds the query with quotes to mitigate against this.


tobi-wan-kenobi:
Makes sense, many thanks!

Issue #641, opened 2020-05-30
Unit tests for modules


tobi-wan-kenobi:
### Feature Request
Right now, the project's unit test coverage is ~15%. The core framework is covered with 90+%, but the module's make up the brunt of the code - and those are barely tested.

Any code that adds tests for modules would be much appreciated.

(Incidentially, also moving over to pytest would be highly appreciated)

rosalogia:
Kinda wanna help out with this... I do need to spend some time looking into good testing practices and pytest first though. I'll try to stay in touch over it!

tobi-wan-kenobi:
Any help on that would be much appreciated!

Fair warning: It's a huge task, as modules make roughly 85% of the code, and we really need a module by module decision what a meaningful test is (ie it must still be standalone, not requite internet access etc.). As such, heavy mocking will be involved, I guess.

Thanks a lot in advance, always glad for people who are up to a challenge :-)

rosalogia:
I appreciate the warning! How about this: I'll look over the modules right now, and if I have any ideas off the top of my head, I'll let you know. If you have the time to be working on this at the same time as me, or at least to be in discussion with me about it, let's find a place to have that continuous conversation. If we can bounce ideas off of one another, we should be able to make progress in test _design_ much faster. Totally understand if that's not something you presently have time for though!

tobi-wan-kenobi:
Yes, I fully agree, that sounds like a very good plan.

Unfortunately, the amount of time I can currently spend on the project is sporadic and at odd hours, so for me, "discussing" offline via the ticket works "best". It also invites others to participate, so that's a plus (I agree with you that real-time discussions would make for much quicker progress, though).

A first step could be to just migrate the existing kernel.py test over to pytests - it also shows how a very very simple module could maybe be mocked.

Thanks!

rosalogia:
I'll take a look and see if I can get started on this today!

rosalogia:
Taking a look at `kernel.py` and there are two things I'm trying to figure out:

If I'm understanding correctly, should `someKernel` and `module` be defined as the return values of functions decorated with `@pytest.fixture`?

I'm trying to understand how the [`mock.patch` line](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/master/tests/modules/test_kernel.py#L15) can be replicated with pytest. Admittedly, I'm not super well read or practiced with mocking or monkey patching, so I'll take some time to read and experiment before making any pull requests.

[...]

PR #693, opened 2020-08-29
Add Network Traffic module tests

izn:
Hey there.

In this PR we create some ~basic~ tests to `network_traffic` module. :tada: 

I tried to cover only the `Widget.full_text` and `Module.update` methods. I believe these are the only methods that the core module uses for the main behavior.

I'm accepting suggestions! :relaxed: 

Related issue: #641.


tobi-wan-kenobi:
Very, very cool, many thanks for the PR!

PR #694, opened 2020-08-30
small improvements in octoprint plugin

bbernhard:
* octoprint sometimes returns additional information when the 3d printer is offline.
  so, it's better to check if the octoprint job state starts with "Offline".

* in case the returned job state is really long, truncate it.


tobi-wan-kenobi:
Looks good, many thanks! Cool module, btw :)

PR #695, opened 2020-08-30
Add memory module tests

izn:
Hey there.

In this PR we create some tests to **memory** module. :tada:
I'm accepting suggestions! :relaxed:

I couldn't test only one behavior because I believe these condicionais are wrong in the module. cc @tobi-wan-kenobi

```python
# OK, here we assume that `MemAvailable` key is optional.
if "MemAvailable" in data:
    used = data["MemTotal"] - data["MemAvailable"]
else:
    used = (
        data["MemTotal"]
        - data["MemFree"]
        - data["Buffers"]
        - data["Cached"]
        - data["Slab"]
    )

self._mem = {
    "total": util.format.byte(data["MemTotal"]),
    # Here we assume this key is required.
    "available": util.format.byte(data["MemAvailable"]), 
    "free": util.format.byte(data["MemFree"]),
    "used": util.format.byte(used),
    "percent": float(used) / float(data["MemTotal"]) * 100.0,
}
```

Related issue: #641.

### DIFF
--- a/tests/modules/core/test_memory.py
+++ b/tests/modules/core/test_memory.py
@@ -1,5 +1,121 @@
 import pytest
+from unittest import TestCase, mock
 
-def test_load_module():
-    __import__("modules.core.memory")
+import core.config
+import core.widget
+import modules.core.memory
 
+def build_module(args = []):
+    config = core.config.Config(args)
+    return modules.core.memory.Module(config=config, theme=None)
+
+def memory_widget(module):
+    return module.widgets()[0]
+
+def meminfo_mock(
+    total,
+    available,
+    free = 0,
+    buffers = 0,
+    cached = 0,
+    slab = 0
+):
+    data = []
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks a lot, looking gokd!

What do you mean vy conditions that are wro g in the code, though?

izn:
I mean, I believe the code is wrong (or deprecated) because we try to call `MemAvailable` even if that key does not exist (I'm not sure if this is possible, but the actual code shows this is possible).

```python
>>> data = { 'MemFree': 1, 'Buffers': 2 }
>>> data['MemAvailable']
  File "<stdin>", line 1, in <module>
KeyError: 'MemAvailable'
>>> data['MemFree']
1
```

izn:
Well, something went wrong with older Python versions. I'm taking a look...

PR #696, opened 2020-08-30
Fix memory module tests

izn:
Hey there.

In this PR we change the way `memory` module parses `/proc/meminfo`.

This is a known bug in older Python versions related to `mock_open`: https://bugs.python.org/issue32933

Related PR: #695


izn:
Taking a look using Python 3.4 :eyes:

tobi-wan-kenobi:
Wow, great dedication to improving the unit tests! Many thanks, this is much appreciated!

PR #701, opened 2020-09-02
Improve network traffic module tests

izn:
Hey there.

This PRs improves `network_traffic` module tests and remove useless `if` conditional from the module logic. :tada:

### DIFF
--- a/bumblebee_status/modules/contrib/network_traffic.py
+++ b/bumblebee_status/modules/contrib/network_traffic.py
@@ -97,9 +97,6 @@ class BandwidthInfo(object):
         """Return default active network adapter"""
         gateway = netifaces.gateways()["default"]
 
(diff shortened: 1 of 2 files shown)


bumblebee_status/modules/contrib/network_traffic.py
```diff
@@ -97,9 +97,6 @@ def default_network_adapter(cls):
         """Return default active network adapter"""
         gateway = netifaces.gateways()["default"]
 
-        if not gateway:
-            raise "No default gateway found"
-
```
izn, line 102: This conditional block is not useful because line #98 explodes if the `default` key is not found. :thinking:

tobi-wan-kenobi:
Looks good, many thanks for another contribution to the test suite!

PR #704, opened 2020-09-03
Add uptime and public ip tests

izn:
Hey there.

This PR adds some `uptime` and `publicip` module tests. :tada:

Related issue: #641.

### DIFF
--- a/tests/modules/contrib/test_publicip.py
+++ b/tests/modules/contrib/test_publicip.py
@@ -35,10 +35,7 @@ class PublicIPTest(TestCase):
 
         assert widget(module).full_text() == 'n/a'
 
-    @mock.patch('util.location.public_ip')
-    def test_interval_seconds(self, public_ip_mock):
-        public_ip_mock.side_effect = Exception
-
+    def test_interval_seconds(self):
         module = build_module()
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
:thumbsup: Very nice, thanks!

btw.: Did you already consider using https://pypi.org/project/pytest-mock/ for mocking? It might ease integration with pytest.

izn:
I will take a look. Tks!

By the way, what do you think about creating a page about writing tests? We can specify which modules we recommend and some test standards.

tobi-wan-kenobi:
Good idea! So far, I thought looking at thecexisting tests is sufficient, but as those grow in size now (thanks!), having a wiki page would be useful. I will try to find time soon to get it started.

If you add someething to the RST structure in docs/, it should qutomatically end up on readthedocs :-)

tobi-wan-kenobi:
@izn I have now updated the documentation somewhat, with at least rudimentary testing guidelines.

Funnily enough, they were present already, just in a non-intuitive location :)

izn:
@tobi-wan-kenobi great! I will take a look soon. ✨

PR #705, opened 2020-09-04
[modules/dunstctl]: Toggle dunst v1.5.0+ notifications using dunstctl

cristianmiranda:
The current `dunst` module kills dunst when toggled which prevents the daemon to execute scripts if configured. This new module is intended to be used only if you have the latest version of dunst (v1.5.0 or greater).

Have a great weekend! 😄

### DIFF
--- a/bumblebee_status/modules/contrib/dunstctl.py
+++ b/bumblebee_status/modules/contrib/dunstctl.py
@@ -1,0 +1,42 @@
+# pylint: disable=C0111,R0903
+
+"""
+Toggle dunst notifications using dunstctl.
+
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Nice, many thanks!

Do you think it would maybe make sense to merge those two into a single module "dunst" and use some logic to autodetect which approach works better?

(I am a big fan of both autodetect when possible and having a single module for one purpose)

cristianmiranda:
Good idea! Let me see what I can do.
I'm also going to open a new PR since I forgot to upload docs screenshots for this module and `thunderbird`.

PR #706, opened 2020-09-05
[modules/thunderbird]: Thunderbird's unread email counts by inbox

cristianmiranda:
I've been using this one for a while now, but never had the time to create a PR. Here it goes 😄 

I have two main inboxes I care about: Personal & Work

### DIFF
--- a/bumblebee_status/modules/contrib/thunderbird.py
+++ b/bumblebee_status/modules/contrib/thunderbird.py
@@ -1,0 +1,89 @@
+# pylint: disable=C0111,R0903
+
+"""
+Displays the unread emails count for one or more Thunderbird inboxes
+
+Parameters:
+    * thunderbird.home: Absolute path of your .thunderbird directory (e.g.: /home/pi/.thunderbird)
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
As a Thunderbird user: Great addition, many thanks!

tobi-wan-kenobi:
What would be a possible addition is some autodetect whether the most common locations ($HOME/.thunderbird, etc) exist, if home is not specified, wuat do you think?

cristianmiranda:
> What would be a possible addition is some autodetect whether the most common locations ($HOME/.thunderbird, etc) exist, if home is not specified, wuat do you think?

I had some trouble using `$HOME/.thunderbird` and `~/.thunderbird`. I'll do some research on how to properly get the current user's home directory so we can use it and remove `thunderbird.home` param.

tobi-wan-kenobi:
@cristianmiranda Yeah, Python is weird in that regards. You have to call `os.path.expanduser()` (I think), so that the path variables are resolved.

Assumedly, that is for security reasons? You can have a look at the code that iterates the theme folders, it makes use of that.

Cheers!

PR #711, opened 2020-09-28
Force feedparser to 6.0.0b1

izn:
Hey ya.

`feedparser` 6.0.0b2+ does not support Python 3.5- anymore.

This PR forces `feedparser` to `6.0.0b1` just to keep things running.


tobi-wan-kenobi:
Ah, good catch, many thanks!

PR #714, opened 2020-10-01
Create symbolic links

### LINKED ISSUE #713: Modules names with hyphens (-)
Hey, ya.

Should we rename the modules below to match the [Python Naming Conventions](https://www.python.org/dev/peps/pep-0008/#package-and-module-names)?

```sh
modules/core/layout-xkb.py
modules/contrib/layout-xkbswitch.py
modules/contrib/arch-update.py
modules/contrib/battery-upower.py
```

These modules work as expected (I think) in the status bar, but I can't import these modules into a test file using `import`.

```sh
E       import modules.contrib.arch-update
E                                  ^
E       SyntaxError: invalid syntax
```

Any ideas?

izn:
Hey!

This PR creates symbolic links to modules with hyphens to match the Python Naming Conventions.

As discussed, we would like to keep the current naming with hyphens as an option because of backwards compatibility.

Fixes #713

### DIFF
--- a/bumblebee_status/modules/contrib/arch_update.py
+++ b/bumblebee_status/modules/contrib/arch_update.py
@@ -1,0 +1,1 @@
+arch-update.py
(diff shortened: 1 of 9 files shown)


tobi-wan-kenobi:
Thanks for also updating the documentation!

PR #717, opened 2020-10-04
Add org-mode TODO module

joshbarrass:
This is based on the existing `todo.txt`-based module, but is modified to better suit the TODO system built into emacs's org-mode. Documents in org-mode can contain TODO lines as well as non-TODO lines, which this module will differentiate between, and org-mode marks completed items with DONE, which this module can also identify.

The module uses regular expressions to find valid TODO and DONE entries (it will ignore whitespace before the asterisks, allow any number of asterisks for deeply embedded TODOs, and will ignore whitespace between the asterisks and the TODO marker itself) and can display it as either (by default) the total number of TODOs completed out of the total number (i.e. 1/4 = 1 TODO completed out of 4) or the total number remaining (i.e. 1/4 = 1 TODO remaining out of 4).

Left clicking the module on the status bar will open the TODO file in emacs.

### DIFF
--- a/bumblebee_status/modules/contrib/todo_org.py
+++ b/bumblebee_status/modules/contrib/todo_org.py
@@ -1,0 +1,57 @@
+# pylint: disable=C0111,R0903
+"""Displays the number of todo items from an org-mode file
+Parameters:
+    * todo_org.file:      File to read TODOs from (defaults to ~/org/todo.org)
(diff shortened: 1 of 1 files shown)


joshbarrass:
Interesting that codeclimate sees it all as TODOs :upside_down_face:

tobi-wan-kenobi:

PR #718, opened 2020-10-04
Add "concise controls" to Spotify module

joshbarrass:
This adds more concise controls to the Spotify module. I remember these being available in the module a while back, but they seem to have been removed in favour of having additional buttons on the status bar. This merge request re-implements these old controls as an alternative to the current ones. They are disabled by default, but, when enabled using the `spotify.concise_controls` option, they allow you to completely remove the other buttons (add `spotify.layout="spotify.song"` to your config) and still retain full control.

The controls are added to the `spotify.song` widget, and are:
 - Left Click: Toggle pause
 - Wheel Up: Next
 - Wheel Down: Previous

I have tried to retain the existing system for registering the inputs by extending the current functionality to accept a list of `callback_options`, though I am unsure of the reason for doing it this way instead of just registering the inputs directly. If anyone knows and could explain the reasoning, I would appreciate it.


tobi-wan-kenobi:
I think the callbacks option was/is mostly a way to handle callbacks for dynamic widget lists (i.e. prev/next might or might not be present).

Thanks a lot for the PR!

PR #728, opened 2020-10-18
Provide alternative dunstctl implementation

joachimmathes:
Hey @tobi-wan-kenobi,

I realize that a `dunstctl` module has already been provided. So, why this PR?
TL;DR: I didn't check the main branch. :roll_eyes:

I'm currently running the latest bumblebee-status version 2.0.5 on my system (since its production I stick to dedicated version tags instead of the main branch), but I wanted the `dunst` module to make use of the `dunsctl` cli, which was introduced in dunst v1.5.0. So, I started to implement a new `dunstctl` module, with the intent to provide a PR, but apparently without checking main first :roll_eyes:. When I was finally preparing my implementation for a PR, I realized that @cristianmiranda has already provided this feature. I could have deleted my code and just switched to main, but I decided to let the project maintainers decide, if my implementation is worth a look. :slightly_smiling_face: Thus, my effort is not in vain and since my approach is a little bit different some value might be added. The implementation
* is avoiding to cache the current state by making use of the dunstctl toggle functionality
* introduces a third state _unknown_, if dunst is not running.

I know that there is a test for `dunstctl` module, which I will adjust, of course, depending on your decision. Please let me know what you think.

### DIFF
--- a/bumblebee_status/modules/contrib/dunstctl.py
+++ b/bumblebee_status/modules/contrib/dunstctl.py
@@ -1,43 +1,42 @@
 # pylint: disable=C0111,R0903
 
-"""
-Toggle dunst notifications using dunstctl.
+"""Toggle dunst notifications using dunstctl.
 
-When notifications are paused using this module dunst doesn't get killed and you'll keep getting notifications on the background that will be displayed when unpausing.
-This is specially useful if you're using dunst's scripting (https://wiki.archlinux.org/index.php/Dunst#Scripting), which requires dunst to be running. Scripts will be executed when dunst gets unpaused.
+When notifications are paused using this module dunst doesn't get killed and
+you'll keep getting notifications on the background that will be displayed when
+unpausing. This is specially useful if you're using dunst's scripting
+(https://wiki.archlinux.org/index.php/Dunst#Scripting), which requires dunst to
+be running. Scripts will be executed when dunst gets unpaused.
 
 Requires:
     * dunst v1.5.0+
 
 contributed by `cristianmiranda <https://github.com/cristianmiranda>`_ - many thanks!
+contributed by `joachimmathes <https://github.com/joachimmathes>`_ - many thanks!
 """
 
 import core.module
 import core.widget
 import core.input
-
 import util.cli
 
 
 class Module(core.module.Module):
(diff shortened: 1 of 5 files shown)


tobi-wan-kenobi:
Thanks for the PR! To me, your implementation looks fine, I particulary like the simplifications introduced by using toggle.

I will wait a while to see whether @cristianmiranda wants to chime in, but I lean towards mergong your PR.

Just one note: Please do not delete the attribution and shorten the help text so much (except to better reflect what the module's purpose is). Instead, please just add yourself to the list of contributors at the bottom of the docstring.

Thanks a lot!

cristianmiranda:
Hey, I love @joachimmathes's contribution. I agree with @tobi-wan-kenobi on the help text, perhaps something like this:

```
"""
Toggle dunst notifications using dunstctl.

When notifications are paused using this module dunst doesn't get killed and you'll keep getting notifications on the background that will be displayed when unpausing.
This is specially useful if you're using dunst's scripting (https://wiki.archlinux.org/index.php/Dunst#Scripting), which requires dunst to be running. Scripts will be executed when dunst gets unpaused.

Requires:
    * dunst v1.5.0+

contributed by `cristianmiranda <https://github.com/cristianmiranda>`_ - many thanks!
contributed by `joachimmathes <https://github.com/joachimmathes>`_ - many thanks!
"""
```

Thank you both!

cristianmiranda:
Also, I was thinking, we could add another screenshot to the docs of the new "unknown" state of the widget. What do you think?.

joachimmathes:
Thank you, for your fast and nice reply.

The help text only seems shortened, since I didn't know of the already existing `dunstctl` and I wanted to wait for your reply, before I continue my work. Plus, I was not sure about thanking myself. :smile:

 @cristianmiranda I will apply your help text suggestion. Unfortunately, I can not help with the screenshots, since I seem to miss some glyphs in the font. (This is currently not an issue for me, because I prefer plain ascii theme anyway.)

I will also take care of the module test.

joachimmathes:
* Tests are adjusted.
* Help text is extended.

tobi-wan-kenobi:
Looks good to me, many thanks for the contribution!

PR #729, opened 2020-10-26
Fix arch-updates off-by-one.

spxtr:
There's a newline in the output so this overcounts by one.


tobi-wan-kenobi:
Good catch, many thanks for the PR!

PR #730, opened 2020-10-26
update python name in shebang to respect PEP-0394

donfranio:
If python2 is not available anymore there might be the case, that no python is available with the name {{python}}, eg. on debian bullseye.


tobi-wan-kenobi:
Looks good to me, thanks!

PR #733, opened 2020-11-06
[module] Improved smartstatus with combined_singles

martindoublem:
Added combined_singles as way to detect the drives that are permanently inside your machine and not plugged via USB.

As USB flash drives without smartstatus caused the module to crash.


tobi-wan-kenobi:
very nice addition, thanks!

PR #736, opened 2020-11-24
add rofication module

jebaum:
I switched from dunst to rofication to manage notifications on my machine and missed the dunst indicator in bumblebee-status. Figured I'd add one


tobi-wan-kenobi:

PR #738, opened 2020-11-29
Add mask for avahi duplicate interfaces

cdbrkfxrpt:
Hi, me again,

I don't know if this a pervasive issue, but I've found that for some reason (haven't investigated it particularly much, tbh) interfaces are sometimes duplicated and amended with `:avahi`, which this now masks out.

Cheers

### DIFF
--- a/bumblebee_status/modules/core/nic.py
+++ b/bumblebee_status/modules/core/nic.py
@@ -93,7 +93,8 @@ class Module(core.module.Module):
     def _update_widgets(self, widgets):
         self.clear_widgets()
         interfaces = [
-            i for i in netifaces.interfaces() if not i.startswith(self._exclude)
+            i for i in netifaces.interfaces()
+            if not i.startswith(self._exclude) and ":" not in i
         ]
         interfaces.extend([i for i in netifaces.interfaces() if i in self._include])
 


tobi-wan-kenobi:
Thanks for the PR!

Would you be amendable to adding a ":avahi" pattern to the default exclude list, instead?

If I understand the diff correctly, this would remove any interface with a : in it from the list of shown interfaces, and in a way that is hard for the user to circumvent.

I'd extend the exclude list to be a pattern, in that case, if that sounds fine to you.

cdbrkfxrpt:
You understand the diff correctly; that's what it does. I had thought of what you're suggesting but decided that with `:` being something not commonly found in interface names outside of netifaces output for duplicated devices, it would be best to exclude any interface name containing that. Particularly since the error is also fairly obscure; bumblebee-status shows SSID and IP address correctly, but also dumps the raw error message coming out of `iwgetid` into the status line.

Feel free to amending the PR as you see fit, though. I haven't put that much thought into it and am doing a fair bit of assuming stuff™ as well. If you think excluding just `:avahi` that's completely fair.

tobi-wan-kenobi:
Thanks! I did some research and it seems : is generally used for subimterfaces (e.g. two IPs on the same interface)

I will try to come up wih something asap to fix your issues.

tobi-wan-kenobi:
@cdbrkfxrpt Thanks for your patience! Could you please tell me whether your issue is fixed now? The devices should now be gone by default.

cdbrkfxrpt:
Thanks for working on this!

With the current HEAD on main I get a permissions error on `iw dev <iface> link`. I've solved it, and with the fix this here issue is then also resolved.

I'm opening another PR for the other fix.

PR #739, opened 2020-11-29
add python 3.9 support

eumiro:
And please consider dropping Python 3.4 and 3.5 support (they are after EOL) to allow the features of Python 3.6+.


tobi-wan-kenobi:
Many thanks!

And fair enough, I will read up on that and drop support, umless there are good reasons not to.

PR #740, opened 2020-11-30
Change iw call in module nic from link to info

cdbrkfxrpt:
On my machine, the call `iw dev <iface> link` fails due to permissions. However, `iw dev <iface> info` does not require user privileges and has similar output.

This change switches to `iw dev <iface> info` and adjusts the information retrieval from the output of that command.


cdbrkfxrpt:
It seems code climate doesn't like regular expressions becoming longer than 5 characters? Welp

tobi-wan-kenobi:
looks good, many thanks for the PR!

PR #742, opened 2020-12-02
Add TiB to disk units, add SI unit option for disk space

cdbrkfxrpt:
Hi, me again ^^

I think adding in TiB is a no-brainer but on the "to SI or not to SI" front, as well as this rather naive implementation, I'm definitely open to suggestions.

Also, in case you like this, let me know if there's anything else I need to do for the information re unit system to end up in the docs.

Cheers!


tobi-wan-kenobi:
Looks good to me, many thanks for the addition!

PR #751, opened 2020-12-16
Clarify the interval values.

P-Storm:
I saw that there was some ambiguity between the interval. I tried to enter a value '1' for myself and that seemed to fail.

`Version: bumblebee-status  2.1.1-1 (AUR)`

Maybe add a fallback if it is only a number. For now this clarification in the documents was enough for me :)

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -50,11 +50,11 @@ entering:
    $ ./bumblebee-status -l themes
    $ ./bumblebee-status -l modules
 
-To change the update interval, use:
+To change the update interval, use a value like ('5s' or '1m') use:
 
 .. code-block:: bash
 
-   $ ./bumblebee-status -m <list of modules> -p interval=<interval in seconds>
+   $ ./bumblebee-status -m <list of modules> -p interval=5s
 
 The update interval can also be changed on a per-module basis, like
 this:


tobi-wan-kenobi:
Oh. Using a simple integer *should* work (and count as seconds)

What happened when you tried that?

P-Storm:
Well, for unrelated reasons I had to reinstall my distro, and wanted to use the global 1 seconds interval.

It looks like `-p interval=1` (and also `interval=1s`) is not working. When changing it to time.interval=1 it works as intended. (same as time.interval=1s)

So it is a bug report for the interval, and maybe a rewording that you can use a suffix if you want, otherwise it will defaults to seconds.

tobi-wan-kenobi:
Ah, I see the confusion now. That indeed needs more explanation.

The issue is that the "global" interval even defaults to 1 (meaning 1s), but that other modules, such as the `time` module can override that default and by default, update less regularly.

I.e. you could think of the global interval as some kind of "heartbeat", the highest resolution interval that *any* of the other modules support. It is a bit difficult to document that succinctly, though. I will try to come up with something.

Thanks for your efforts!

tobi-wan-kenobi:
I added some clarification, and would appreciate it a lot if you could let me know whether they look fine for you.

Thanks!

P-Storm:
Yes, that works for me. Only thing that I'm missing now is, what is the default 'global' interval? I'm guessing 1s, but maybe add it somewhere. Other than that, that looks like a great addition. (And maybe add some reference to the advanced page, because I would expect such a thing also in the advanced. Maybe move the explanation to the advanced page and refer it from the introduction for more details.

It is clear for me what the variable does now :)

PR #752, opened 2020-12-18
[modules/spotify] Add initial spotifyd compatibility

### LINKED ISSUE #750: Is it possible to run spotifyd with bumblebee-status ?
Hello! 
How would I make status bar to read spotifyd? That would be so awesome!

gkeep:
Adds [spotifyd](https://github.com/Spotifyd/spotifyd) compatibility to `spotify` module. Closes #750.

To enable compatibility set `spotify.bus_name` parameter to `spotifyd` in your config. The parameter will default to `spotify` unless set.

One small issue: the module sometimes disappears. Very likely `spotifyd`'s fault. (Sometimes does not return metadata)


tobi-wan-kenobi:
Looks good, thanks!

One note: I'd recommend to directly use the variable in the dbus id strings usig format strings (i.e. "org.media.whatever.{}".format(self._bus_type))

That gets rid of a lot of if/else and makes the module even more extensible.

If you are worried about uncontrolled input, you could sanitize the value in __init__.

PR #754, opened 2020-12-27
[modules/playerctl] Add format and layout parameters

gkeep:
This PR adds format and layout parameters similar to `spotify` module (https://github.com/tobi-wan-kenobi/bumblebee-status/issues/750#issuecomment-748465878).

Right now, the module calls to the `playerctl` shell program via `util.cli.execute`, which can be slow and unreliable.

I wanted to rewrite the module to support the [playerctl API library](https://github.com/altdesktop/playerctl#using-the-library), but I found the documentation incredibly lacking and not enough to write something functional with it (if you feel otherwise, please use it).

### DIFF
--- a/bumblebee_status/modules/contrib/playerctl.py
+++ b/bumblebee_status/modules/contrib/playerctl.py
@@ -5,57 +5,113 @@
 Requires the following executable:
     * playerctl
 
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
many thanks for the improvement!

PR #766, opened 2021-02-26
Add code to enable scrolling of shell module output

michalroxorpl:
In scenarios requiring use of own shell modules it is impossible to scroll wide output produced by shell module. Let's consider the following scenario:
```
[core]
modules=shell:crypto

[module-parameters]
# shell:crypto
crypto.command=bash -c 'printf "=%.2f$ XRP=%.2f$ ADA=%.2f$" $(curl -s rate.sx/1BTC) $(curl -s rate.sx/1XRP) $(curl -s rate.sx/1ADA)'
crypto.async=true
crypto.scrolling.width=15
crypto.scrolling.bounce=true
```

Without the modification proposed the following static full-width shell module output will be produced:
Selection_330

With the modification proposed the following narrowed bouncing / scrolling output is possible:
Selection_331

### DIFF
--- a/bumblebee_status/modules/contrib/shell.py
+++ b/bumblebee_status/modules/contrib/shell.py
@@ -54,6 +54,7 @@ class Module(core.module.Module):
     def set_output(self, value):
         self.__output = value
 
+    @core.decorators.scrollable
     def get_output(self, _):
         return self.__output
 


tobi-wan-kenobi:
Ah, yes. Makes a lot of sense, many thanks for the contribution!

michalroxorpl:
Thank you for accepting my requests so fast.

PR #775, opened 2021-03-17
Set default scrolling width to match text length

es80:
Making the output of the 'shell' module scrollable led to the introduction of a good deal of white space for my setup since the text I output is quite short.

I found that I was able to correct this by either setting the parameter `shell.scrolling.width` to the exact width of my output, or by setting `shell.scrolling.makewide` to `False`.

I wonder if you might consider making one of these the default setting for the parameter. It might make it less confusing for any newcomers trying to understand where the white space is coming from if their output is of length less than 30.

### DIFF
--- a/bumblebee_status/core/decorators.py
+++ b/bumblebee_status/core/decorators.py
@@ -74,7 +74,7 @@ def scrollable(func):
             widget.set("scrolling.direction", "right")
         widget.set("__content__", text)
 
-        width = util.format.asint(module.parameter("scrolling.width", 30))
+        width = util.format.asint(module.parameter("scrolling.width", len(text)))
         if util.format.asbool(module.parameter("scrolling.makewide", True)):
             widget.set("theme.minwidth", "A" * width)
         if width < 0 or len(text) <= width:


tobi-wan-kenobi:
good point, I did not think about/test that!

the proposed change, however, will break the music players (they would change size after each song) - the scrolling was initially implemented to and geared towards those modules.

How about if we introduce a way to change the defaults for specific modules, eg. the shell module?

btw - once this is fixed, I think it warrants a bugfix release :)

thanks for the report!

es80:
Ah, I did wonder if that might break someone else's scrollable text.

How about something like this in the `shell` module?

```diff
diff --git a/bumblebee_status/modules/contrib/shell.py b/bumblebee_status/modules/contrib/shell.py
index cb7bc38..608ea95 100644
--- a/bumblebee_status/modules/contrib/shell.py
+++ b/bumblebee_status/modules/contrib/shell.py
@@ -47,12 +47,17 @@ class Module(core.module.Module):
             self.__output = "please wait..."
             self.__current_thread = threading.Thread()
 
+        if self.parameter("scrolling.makewide") is None:
+            self.set("scrolling.makewide", False)
+
         # LMB and RMB will update output regardless of timer
         core.input.register(self, button=core.input.LEFT_MOUSE, cmd=self.update)
         core.input.register(self, button=core.input.RIGHT_MOUSE, cmd=self.update)
 
     def set_output(self, value):
         self.__output = value
+        if self.parameter("scrolling.width") is None:
+            self.set("scrolling.width", len(value))
 
     @core.decorators.scrollable
     def get_output(self, _):
@@ -61,7 +66,7 @@ class Module(core.module.Module):
     def update(self):
         # if requested then run not async version and just execute command in this thread
         if not self.__async:
-            self.__output = util.cli.execute(self.__command, shell=True, ignore_errors=True).strip()
+            self.set_output(util.cli.execute(self.__command, shell=True, ignore_errors=True).strip())
             return
 
         # if previous thread didn't end yet then don't do anything

```

Was that what you had in mind? 

Not sure which of `makewide` or `width` is better to add a default for. I figured `makewide` could go in the constructor but `width` might need to change with each update.

tobi-wan-kenobi:
Yes, that is exactly what I was looking for.

I think if makewide is false, you do nor need to set maxwidth, it is just when the output will start scrolling (and this covrrs the usecase of the original pr, i brlieve)

tobi-wan-kenobi:
Hope this change works for you!

es80:
Excellent! Thanks again for being so quick to look at this.

PR #778, opened 2021-03-27
[themes] add albiceleste-powerline and rastafari-powerline

c05m4r:


tobi-wan-kenobi:
very nice themes, many thanks!

PR #789, opened 2021-05-10
Fix parameters name for the sun module

fredj:


tobi-wan-kenobi:
That is an embarrassing one, thanks for fixing!

PR #793, opened 2021-06-11
[modules/playerctl]: use `playerctl -f` and add `playerctl.args`

### LINKED ISSUE #767: playerctl not displaying all information
#### Description
Affected module: playerctl
Version used: latest git

Only displays previous and next icons for output from Vlc. Other icons don't display. Pause/resume works clicking on the blank space between the buttons. Current song does not display. No right mouse functionality, whether there is supposed to be any. The icons are in the awesome-fonts.json.

#### How to reproduce
simply run it. Haven't tried with other players since none are installed.

ouuan:
1. Use `playerctl -f` to format, which is more powerful. This also fixes #767, which is caused by missing a few fields of the metadata.
2. Add `playerctl.args`, so that users can choose a specific player, etc.
3. Display nothing when there's no running player.

This is a breaking change. Users need to change `{title}` to `{{title}}`.

I'm not sure how should I handle a breaking change. Should I add a new parameter like `playerctl.playerctl_format`, or replace `{artist}` with `{{artist}}`, or just break it?

### DIFF
--- a/bumblebee_status/modules/contrib/playerctl.py
+++ b/bumblebee_status/modules/contrib/playerctl.py
@@ -6,12 +6,14 @@ Requires the following executable:
     * playerctl
 
 Parameters:
-    * playerctl.format:   Format string (defaults to '{artist} - {title}')
-      Available values are: {album}, {title}, {artist}, {trackNumber}
+    * playerctl.format:   Format string (defaults to '{{artist}} - {{title}}  {{duration(position)}}/{{duration(mpris:length)}}')
+      The format string is passed to 'playerctl -f' as an argument. See the 'Format Strings' section of 'man playerctl' for more information.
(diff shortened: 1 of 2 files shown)


gkeep:
Looks good to me!

I think the best option is to just break it. 

The change to parameters will be noted in the changelog for a future release. Plus users who didn't notice the change will notice it, when the module doesn't work after the update.

tobi-wan-kenobi:
> The change to parameters will be noted in the changelog for a future release

I am grateful for that vote of confidence, now I just need to jot that down so I don't forget :P

I agree, I would also just go ahead and break the interface. *If* somebody complains, we can always re-introduce a compatibility layer. But generally, I would like to avoid code growth because we try to stay backwards compatible always.

tobi-wan-kenobi:
+ the diff looks good to me, so in it goes. Thanks a lot for the contribution!

PR #795, opened 2021-06-24
Add active gpu module using optimus-manager

sayansil:
I was searching for a little module that could show me my active GPU (intel/nvidia) but ended up making one by myself after failing to find one. For those who use optimus-manager to switch between GPUs, this helps determine the active one.

Looks something like: 
image


tobi-wan-kenobi:
Very nice, many thanks! I will merge the PR as is, but for future improvements: In the util.cli library, you can find nice high level abstractions from subprocess.Popen.

PR #798, opened 2021-06-30
Use the existing util.cli module in contrib/optman 

sayansil:
Thanks for the suggestion last time!


tobi-wan-kenobi:
Looks good, thanks a lot!

PR #800, opened 2021-07-09
Network

nepoz:
Network is a simple module to show user's default network interface information. Includes SSID and signal strength for wireless connections.


tobi-wan-kenobi:

I have to admit that I do not understand the core difference between this module and 'nic', but since it probably is something I am overlooking, and the code looks very well written, I will just go ahead and merge it.

Thanks!

nepoz:
Thanks for accepting my contribution! Figured it might be a little simpler/more obvious than nic is all.

PR #807, opened 2021-08-10
Dependency for powerline themes

soykan:


tobi-wan-kenobi:
Thanks a lot for the addition!

PR #810, opened 2021-08-12
Improving docs

soykan:
I added tkinter as a dependency in requirements.
If tkinter is not installed with python3 like my system the module gives this error "unable to load module"


tobi-wan-kenobi:
Thanks a lot! Doc improvements are always highly welcome :)

PR #806, opened 2021-07-24
feat: add GPU usage % and GPU memory usage % to nvidiagpu

tomsaleeba:
Adding support for these two values:
```console
$ nvidia-smi -q | grep -A 2 Utilization
    Utilization
        Gpu                               : 4 %
        Memory                            : 21 %
```
...into the format string for the `nvidiagpu` module.

:heart_eyes: thanks for the awesome tool.

PR #813, opened 2021-08-16
fix: correct mem usage to be mem *io* usage

tomsaleeba:
I realised I made a mistake in [my previous pr](https://github.com/tobi-wan-kenobi/bumblebee-status/pull/806). The memory percentage value that I added was *not* the used/total usage of the memory itself. Instead it was actually the memory IO time :man_facepalming:. I should've RTFM first.

This PR fixes that mistake and renames the value to what it actually is. Plus it adds the "used/total as a percentage" value that I wanted all along.

Sorry for the mixup :grimacing:


tobi-wan-kenobi:
Thanks a lot for the module, and for the fix!

PR #820, opened 2021-09-20
Introduce threshold dependent shell output coloring with unit present…

michalroxorpl:
…ation

Let's assume we want to have visual alerts (background color change) depending on specific measurable value coming as a result of shell action. This patch makes this possible:

- Call the module

```
[core]
modules=...,shell:temperature,...
```

- Configure the module (the trigger is to define **unit**)

```
# shell:temperature
temperature.command=bash -c 'echo $(cat /sys/class/thermal/thermal_zone0/temp | cut -c 1-2)'
temperature.unit=°C
temperature.warning=75
temperature.critical=85
```

The possible results:
image
image
image

There might be better ideas to solve this requirement and for sure better ways to implement it. If there would be any comments and / or suggestions to make it better, I am open to any critics (even non constructive).

### DIFF
--- a/bumblebee_status/modules/contrib/shell.py
+++ b/bumblebee_status/modules/contrib/shell.py
@@ -40,6 +40,7 @@ class Module(core.module.Module):
     def __init__(self, config, theme):
         super().__init__(config, theme, core.widget.Widget(self.get_output))
 
+        self.__unit= self.parameter("unit", 'echo "no unit configured"')
         self.__command = self.parameter("command", 'echo "no command configured"')
         self.__async = util.format.asbool(self.parameter("async"))
 
@@ -55,7 +56,7 @@ class Module(core.module.Module):
         self.__output = value
 
     def get_output(self, _):
-        return self.__output
+        return self.__output + str(self.parameter("unit"))
 
     def update(self):
         # if requested then run not async version and just execute command in this thread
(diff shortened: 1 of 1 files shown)


Tonus1:
Seems great to me. Looking forward it inclusion in next release.

bumblebee_status/modules/contrib/shell.py
```diff
@@ -40,6 +40,7 @@ class Module(core.module.Module):
     def __init__(self, config, theme):
         super().__init__(config, theme, core.widget.Widget(self.get_output))
 
+        self.__unit= self.parameter("unit", 'echo "no unit configured"')
```
tobi-wan-kenobi, line 43: I think you don't want to use "echo..." as default (this is done in the command to provide a meaningful default command)

Probably using None would be best, as you are checking for that later. Or possibly an empty string to make the concatenation work nice in the output.

bumblebee_status/modules/contrib/shell.py
```diff
@@ -77,8 +78,25 @@ def update(self):
         self.__current_thread.start()
 
     def state(self, _):
-        if self.__output == "no command configured":
-            return "warning"
+        if self.parameter("unit") is None:
```
tobi-wan-kenobi, line 81: I think this can be shortened by using the "threshold_state()" convenoenxe method that is part of the Module class.

tobi-wan-kenobi (commented):
Looks good, but I think adressing the 2 comments I left would be great. Many thanks for the controbution!

tobi-wan-kenobi:
Merging the PR "as is", since it is functionally fine, affects only a single module, and can easily be refactored to incorporate the above suggestions.

Thanks for the contribution!

PR #823, opened 2021-10-11
Create dunst tests

izn:
Hey ya.

This PRs creates some **dunst** tests. :tada:

Related issue: #641.

By the way, should we keep supporting two dunst modules (dunst & dunstctl)?


tobi-wan-kenobi:
Nice!!! Many thanks!

Yeah, that is a good point. I am mostly keeping both versions for backwards compatibility.

PR #830, opened 2021-10-26
added new theme nord-colorful

kushagraa-j:
I was missing the nord colors in the bar so i just used the nord palate to create a new theme. I already use it on my setup.


tobi-wan-kenobi:

kushagraa-j:
:)

PR #838, opened 2021-12-17
[contrib/playerctl]: don't log when no player is found

ouuan:
`playerctl status` returns 1 when no player is found, which caused contrib/playerctl to log many times when there's no player.

### DIFF
--- a/bumblebee_status/modules/contrib/playerctl.py
+++ b/bumblebee_status/modules/contrib/playerctl.py
@@ -85,7 +85,9 @@ class Module(core.module.Module):
 
     def update(self):
         try:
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks a lot for the improvement!

Alternatively, cli.util.execute() has a parameter to ignore errors (ignore_errors=True), but your solutions works great, as well.

ouuan:
> Alternatively, cli.util.execute() has a parameter to ignore errors (ignore_errors=True), but your solutions works great, as well.

`cli.util.execute` with both `ignore_errors=True` and `include_stderr=True` doesn't return `stderr` when exit code is not zero. Is it expected?

Issue #839, opened 2021-12-28
Spotify module writes large amounts of logs to ~/.xsession-errors if spotify is not running


LokiLuciferase:
#### Description
Affected module: spotify
Version used: latest git
Distro/Kernel: Linux Mint 20.2, 5.11.0-43
i3 version: i3-gaps 4.19.1

Hi,

First I'd like to say thank you for this awesome piece of software, I really enjoy having a well-designed hackable status bar written in python.

Secondly, I have been experiencing the following issue:
On my system, the spotify module writes about 2kb/sec of logs to ~/.xsession-errors (which is obviously unsustainable).

These logs have the following shape:

```
...
[2021-12-28 00:37:02,192] spotify          ERROR    org.freedesktop.DBus.Error.ServiceUnknown: The name org.mpris.MediaPlayer2.spotify was not provided by any .service files
Traceback (most recent call last):
  File "/home/testuser/.local/lib/python3.9/site-packages/dbus/bus.py", line 177, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/home/testuser/.local/lib/python3.9/site-packages/dbus/bus.py", line 361, in get_name_owner
    return self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
  File "/home/testuser/.local/lib/python3.9/site-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: Could not get owner of name 'org.mpris.MediaPlayer2.spotify': no such name

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/testuser/.local/lib/python3.9/site-packages/bumblebee_status/modules/contrib/spotify.py", line 134, in update
    self.__get_song()
  File "/home/testuser/.local/lib/python3.9/site-packages/bumblebee_status/modules/contrib/spotify.py", line 120, in __get_song
    spotify = bus.get_object(
  File "/home/testuser/.local/lib/python3.9/site-packages/dbus/bus.py", line 241, in get_object
    return self.ProxyObjectClass(self, bus_name, object_path,
  File "/home/testuser/.local/lib/python3.9/site-packages/dbus/proxies.py", line 250, in __init__
    self._named_service = conn.activate_name_owner(bus_name)
[...]

PR #840, opened 2021-12-28
Silence exceptions in the spotify module

LokiLuciferase:
In reference to: #839

### DIFF
--- a/bumblebee_status/modules/contrib/spotify.py
+++ b/bumblebee_status/modules/contrib/spotify.py
@@ -159,7 +159,6 @@ class Module(core.module.Module):
                     widget.full_text(self.__song)
 
         except Exception as e:
-            logging.exception(e)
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
I think the except block might require a "pass" now, since it's empty (?)

But so ce it works for you, I will go ahead and merge as is.

LokiLuciferase:
> I think the except block might require a "pass" now, since it's empty (?)

The except block still assigns the song string to '', so that should be fine as is (though I just realized that the github app doesn't seem to render the indentation correctly in the changed file view, so that may have been the source of confusion...).

Thanks for the quick response!

tobi-wan-kenobi:
> though I just realized that the github app doesn't seem to render the indentation correctly in the changed file view

That was the issue exactly...

PR #841, opened 2022-01-05
Fixes some QA

fedeliallalinea:
- Excluding the tests folder from the installation
- Change deprecated dash-separated with underscore in setup.cfg


tobi-wan-kenobi:
Many thanks for the improvement!

PR #842, opened 2022-01-06
New emerge status module

fedeliallalinea:
Display information about the currently running emerge process (gentoo linux).


tobi-wan-kenobi:
Looks great, many thanks!

PR #851, opened 2022-02-12
[modules/nic] Added strength indicator for wifi signals

deadbeef2000:
Added the WiFi signal strength in percent and added the warning/critical states to the module depending on the signal strength.
Also lowered the auto-refresh time from 10 to 5 seconds

Warning if < 50
Critical if < 30

(Can be turned off by setting the both limits to 0)

This is my first pull request, if there are any issues or if I forgot anything, don't hesitate to tell me about it ;)


tobi-wan-kenobi:
Looks good, many thanks for the addition!

PR #854, opened 2022-02-14
Fix publicip

### LINKED ISSUE #853: publicip prints "SPEC_VIOLATION: full_text is NULL!"
#### Description
Affected module: `modules/contrib/publicip.py`
Version used: latest (git)

If the Internet connection is down and the publicip module is running, "SPEC_VIOLATION: full_text is NULL!" is printed instead of "n/a".

image

#### How to reproduce

1. Disable the Internet connection
2. Reload bumblebee-status

mihaimorariu:
Fixes #853 

Calling `__data.get(name, default)` (in the `__get` method of `util/location.py`) results in no exception being raised if a key is not available (such as when an Internet connection is down and `__data` does not get filled). In this case, the `update` method of _publicip_ will not execute the `except` branch, but return `None` instead.

The _sun_ module is affected by this bug as well.

### DIFF
--- a/bumblebee_status/modules/contrib/sun.py
+++ b/bumblebee_status/modules/contrib/sun.py
@@ -39,7 +39,11 @@ class Module(core.module.Module):
         self.__sun = None
 
         if not lat or not lon:
-            lat, lon = util.location.coordinates()
+            try:
+                lat, lon = util.location.coordinates()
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Looks good to me, many thanks for the fix!

PR #856, opened 2022-02-20
fix case of Kelvin SI unit in redshift widget

LokiLuciferase:
In the `redshift` widget, currently the temperature gets lower-cased (`6500k`). This PR makes sure that the that the SI unit symbol for Kelvin gets displayed correctly (`6500K`).


tobi-wan-kenobi:
Ah, good catch. Many thanks for the contribution!

LokiLuciferase:
Fastest merge in history :grin: thanks

PR #857, opened 2022-02-28
Pactl

bbernhard:
This is just a small little module which displays the current selected default sink allows to select a different one.

### DIFF
--- a/bumblebee_status/modules/contrib/pactl.py
+++ b/bumblebee_status/modules/contrib/pactl.py
@@ -93,7 +93,6 @@ class Module(core.module.Module):
 
     def __sink(self, internal_sink_name):
         for sink in self.__sinks:
-            logging.info(sink.internal_name)
             if internal_sink_name == sink.internal_name:
                 return sink
         return None
--- a/bumblebee_status/modules/contrib/pactl.py
+++ b/bumblebee_status/modules/contrib/pactl.py
@@ -1,0 +1,142 @@
+# pylint: disable=C0111,R0903
+
+""" Displays the current default sink.
+
+    Left click opens a popup menu that lists all available sinks and allows to change the default sink.
+
+    Per default, this module uses the sink names returned by "pactl list sinks short"
+
+    sample output of "pactl list sinks short":
+
+    2	alsa_output.pci-0000_00_1f.3.analog-stereo	module-alsa-card.c	s16le 2ch 44100Hz	SUSPENDED
+    3	alsa_output.usb-Logitech_Logitech_USB_Headset-00.analog-stereo	module-alsa-card.c	s16le 2ch 44100Hz	SUSPENDE
+
+    As "alsa_output.usb-Logitech_Logitech_USB_Headset-00.analog-stereo" is not a particularly nice name, its possible to map the name to more a
+    user friendly name. e.g to map "alsa_output.usb-Logitech_Logitech_USB_Headset-00.analog-stereo" to the name "Headset", add the following
+    bumblebee-status config entry: pactl.alsa_output.usb-Logitech_Logitech_USB_Headset-00.analog-stereo=Headset
+
+    The module also allows to specify individual (unicode) icons for all the sinks. e.g in order to use the icon 🎧 for the
+    "alsa_output.usb-Logitech_Logitech_USB_Headset-00.analog-stereo" sink, add the following bumblebee-status config entry:
+    pactl.icon.alsa_output.usb-Logitech_Logitech_USB_Headset-00.analog-stereo=🎧
+
+    Requirements:
+        * pulseaudio
+        * pactl
+"""
+
+import logging
+import functools
+
+import core.module
+import core.widget
+import core.input
+
+import util.cli
+import util.popup
+
+
+class Sink(object):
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Nice, many thanks!

Would it maybe make sense to unify that with the pulseaudio module and/or use this as a basis for a rewrite of the pulseaudio module? What's your opinion on that?

bbernhard:
I also thought about that, but unfortunately the left & right mouse buttons are already in use there:

> Displays volume and mute status and controls for PulseAudio devices. Use wheel up and down to change volume, left click mutes, right click opens pavucontrol.

I guess we could make the left/right mouse button actions configurable (like different modes), but not sure if that would be intuitive to use then?

tobi-wan-kenobi:
Ah, I see, good point.

After thinking about it quite a lot, I still don't know what would be best. Having different modules for the same purpose sounds confusing as well.

What if we did something like this: Mouse actions can be overridden alteady via parameters. If we extens that so that modules can register custom actions, we could have parameters like `pasink.left-click=menu-popup`

What do you think about that? Would that fit your purpose?

tobi-wan-kenobi:
I will just go ahead & merge it, we can always improve the modules later :)

bbernhard:
> What if we did something like this: Mouse actions can be overridden alteady via parameters. If we extens that so that modules can register custom actions, we could have parameters like pasink.left-click=menu-popup

> What do you think about that? Would that fit your purpose?

I think that would definitely make sense & would fit my purpose! 

At the moment I am struggling a bit to understand how that would be implemented. I haven't checked the sourcecode yet, but if I understood the [documentation](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/main/docs/features.rst#events) correctly then the mouse action overrides are handled purely by the framework, right?

If I  e.g specify `bumblebee-status -p nic.left-click="pavucontrol"` the framework ensures to invoke the `pavucontrol` executable on a left click. So, there's no logic to implement in the module, right? 

I guess if the `menu-popup` should be handled generically by the framework as well, I guess we need to somehow pass the available dropdown menu options along? Or will there be a concrete `menu-popup` handler that needs to be implemented in the module?

tobi-wan-kenobi:
Right, exactly!

I'm thinking along those lines:

If a custom action is defined, first check if a method with that name is defined for the module (via `getattr()`). If so, invoke it, otherwise try to execute a binary 

I **guess** that should be relatively straight-forward to implent (famous last words)

I'll try to give that a go during the weekend.

tobi-wan-kenobi:
@bbernhard Please see #859 

There's a simple example for the `spacer` module that you can try out like this:

`./bumblebee-status -m spacer -p spacer.left-click=update_text` - basically, the event handler should be the name of the method to invoke, and the method gets exactly one parameter: The event.

Does that help? If so, I'd suggest to drop `pactl` again and extend `pulseaudio` instead, if that is fine with you.

bbernhard:
> @bbernhard Please see #859
> 
> There's a simple example for the `spacer` module that you can try out like this:
> 
> `./bumblebee-status -m spacer -p spacer.left-click=update_text` - basically, the event handler should be the name of the method to invoke, and the method gets exactly one parameter: The event.

Looks good - I think that should work :+1:. Many thanks for the fast implementation!

> If so, I'd suggest to drop `pactl` again and extend `pulseaudio` instead, if that is fine with you.

Sounds good to me :+1:

bbernhard:
> @bbernhard Please see #859
> 
> There's a simple example for the `spacer` module that you can try out like this:
> 
> `./bumblebee-status -m spacer -p spacer.left-click=update_text` - basically, the event handler should be the name of the method to invoke, and the method gets exactly one parameter: The event.
> 
> Does that help? If so, I'd suggest to drop `pactl` again and extend `pulseaudio` instead, if that is fine with you.

This is probably a stupid question, but how do I get the `widget` inside the event handler?

I tried something like this:

```
def __on_sink_selected(self, sink_name):
        logging.info("click")

def select_default_device_popup(self, event):
        logging.info("I am here")
        widget = self.widget()

        menu = util.popup.menu()
        menu.add_menuitem(
                    "hello",
                    callback=functools.partial(self.__on_sink_selected, "hello"),
                )
        menu.show(widget)
```

together with the following config parameter: `pasink.left-click=select_default_device_popup`.

Although the event handler gets called (I see the log message in the log file), I do not see any popup appearing. But I am not sure whether `self.widget()` is the correct way to obtain the widget?

tobi-wan-kenobi:
Hm, that looks like it *should* work, particularly since the pulseaudio modules only have a single widget each.

I woll need to look into this tomorrow...

Can you maybe check the values for seld and widget via logging?

Thanks!

tobi-wan-kenobi:
@bbernhard This is actually a case kf bad naming on my end.

"widget" that is passed to the show() method is actually the "event" (because in a sense, the event represents the widget that was clicked)

Hope that works!

bbernhard:
> @bbernhard This is actually a case kf bad naming on my end.
> 
> "widget" that is passed to the show() method is actually the "event" (because in a sense, the event represents the widget that was clicked)
> 
> Hope that works!

aaah, got it! :grinning: 

Works like a charm now - many thanks!

PR #860, opened 2022-03-04
[contrib] added module persian_date

Hame-daani:
Displays the current date and time in Persian(Jalali) Calendar.

### DIFF
--- a/bumblebee_status/modules/contrib/persian_date.py
+++ b/bumblebee_status/modules/contrib/persian_date.py
@@ -1,0 +1,45 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the current date and time in Persian(Jalali) Calendar.
+
+Requires the following python packages:
+    * jdatetime
+
+Parameters:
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thank you very much for this contribution!

Pretty cool :)

I will merge right away, but one question for your consideration: Would it maybe make sense to eventually merge zhat into the existing datetime module? (for example, use jdatetime instead of datetime if a certain parameter is set, or even autodetect that from the system)

Thank you kindly!

Hame-daani:
yeah, I thought the same since 99 percent of the code is duplicated.

I believe the better approach would be, to set a `datetime library` as an attribute in the class. and then the `full_text method` will use this attribute to get the date. and then other modules like the `persian_date module` can subclass this `datetime module` and set their own library to be used.

This way the road will be smoother for future modules since as far as I know there are other datetime libraries for other existing calendars.

I will work on this if you don't mind.

tobi-wan-kenobi:
Sounds great and I would appreciate it a lot!

PR #861, opened 2022-03-07
Pactl revert

bbernhard:
This reverts the `pactl` module. Instead of creating a separate module (which duplicates parts of the `pulseaudio` module), the changes will be implemented into the `pulseaudio` module instead.


tobi-wan-kenobi:
Great, thank you very much!

PR #864, opened 2022-03-14
Add module aur-update

ishaanbhimwal:
Module that shows AUR packages to update.


tobi-wan-kenobi:
Thank you very much! As a yay user myself, I'll use that module for sure :)

In future iterations, it might be useful to expose the AUR query command as parameter.

PR #865, opened 2022-03-14
Fix aur-update showing wrong update number

ishaanbhimwal:
Fixed `aur-update` showing `Update AUR: 1`  even when there are no updates. `yay` exists with exit code `0` even when there are no updates unlike `checkupdates` which exits with `2`. Though it does returns a empty string which can be used to verify the status of updates available.


tobi-wan-kenobi:
:thumbsup: looks good!

PR #868, opened 2022-03-25
Pulseaudio extension

bbernhard:
A few small `pulseaudio` extensions. Please let me know what you think :)


tobi-wan-kenobi:
Great additions, many thanks!

One idea regarding the names of the devices: `pamixer` has a nice mode (`--list-sinks` and `--list-sources`) that has a human-readable output name as well.

That would have the advantage to automatically match what tools such as `pavucontrol` show.

What do you think?

PR #870, opened 2022-04-08
Changed __time_format to self.__time_format.

arivarton:


tobi-wan-kenobi:
Many thanks for the fix, good catch!

PR #871, opened 2022-04-23
Add hide parameter to playerctl

SamTebbs33:
This PR adds the `hide` parameter to playerctl so that its widgets can be hidden when no players are found. Currently it populates the bar with empty widgets when no players are found, which isn't great.

Because two functions now require the playback status, I refactored that functionality out into its own function.


tobi-wan-kenobi:
Thank you for the PR, this is a really nice change.

I noted 2 small things in the review, one of which I want to duplicate here: Having this enabled by default would be great, in my opinion. bumblebee wants to be "sane defaults", and the sane behaviour, I think, is to hide the player if it is inactive.

I will go ahead and merge "as is" and either wait for another PR or do this change myself, as soon as I find the time to do so.

Thank you again for your contribution!

SamTebbs33:
Thanks for merging :) I didn't make it hide by default so that existing behaviour would be retained, but do personally agree that it makes sense to hide by default.

PR #873, opened 2022-05-01
arch-update: sleep 1 sec before checking

alexcoder04:
When waking up from suspend, there is sometimes a delay connecting to the network, so arch-update gives an error


tobi-wan-kenobi:
That looks like a really good change, thanks! (not using suspend myself, so I never noticed)

Going forward, it might be nice to make the delay configurable, but let's see if there is demand first.

PR #874, opened 2022-05-11
Fix logout item using hardcoded command

Timoses:
Command was hard coded in instead of using the configurable parameter.


tobi-wan-kenobi:
Oh, that is a good find, many thanks for the fix!

PR #875, opened 2022-05-13
Hide progress module if it's inactive

SamTebbs33:
This PR hides the progress module if it is inactive. This will break existing behaviour as it currently displays placeholder text when inactive, so I can make this new behaviour configurable if desired.


Tonus1:
To hide seems the way to go for me. Thank you

tobi-wan-kenobi:
I agree that this is sensible default behaviour.
Thank you very much for your contribution!

SamTebbs33:
Thank you both

PR #876, opened 2022-05-20
Add popup command parameter to the system module

SamTebbs33:
This PR adds the `popupcmd` parameter to the system module, allowing people to override what happens when the module is clicked. I find this useful to run a rofi menu instead of the default menu.


tobi-wan-kenobi:
Ah, for using something like rofi, this is a really nice idea.

Thanks for the contribution!

PR #877, opened 2022-05-26
[modules]: introduce pamixer module

FraSharp:
Signed-off-by: FraSharp <s23265@iisve.it>

### DIFF
--- a/bumblebee_status/modules/contrib/pamixer.py
+++ b/bumblebee_status/modules/contrib/pamixer.py
@@ -1,0 +1,88 @@
+"""get volume level or control it
+
+Requires the following executable:
+    * pamixer
+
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Looks good, many thanks!
I have to admit I am *slightly* concerned about the proliferation of various different modules doing essentially the same thing (P.A. volume control), but I see the point of this module, like the simplicity of the code & ultimately believe in giving users a choice.

Thank you very much for your contribution!

SamTebbs33:
Perhaps extracting out common code and deriving individual modules from that would be good for reducing duplication. If be happy to do that but I don't have any experience with object orientated programming in python.

Issue #888, opened 2022-06-15
awesome-fonts iconset spacing is too wide


anarcat:
### Discussed in https://github.com/tobi-wan-kenobi/bumblebee-status/discussions/887

<div type='discussions-op-text'>

<sup>Originally posted by **anarcat** June 13, 2022</sup>
hi!

so i'm pretty happy with the srcery theme i just made (https://github.com/tobi-wan-kenobi/bumblebee-status/pull/885, thanks for the merge!), but i have a question...

right now, i use the "awesome-fonts" icon set, and it looks like this:

![snap-20220613T165237](https://user-images.githubusercontent.com/796623/173443383-2a6f154c-0b81-4b61-8323-de6e9c4aa363.png)

... which is pretty cool, but it's wiiiide.... the ascii iconset is not much better:

![snap-20220613T165332](https://user-images.githubusercontent.com/796623/173443544-4215aa86-ce00-44ca-a9e8-6292a3750944.png)

... in fact, it's *wider*!

if i remove the iconset altogether, everything gets cramped together, which is probably worse, but at least it uses space more... effectively:

![snap-20220613T165429](https://user-images.githubusercontent.com/796623/173443702-b2415c93-3fea-4a8b-bd8e-17803d93cc7d.png)

... is there a way to find a compromise here? i'd love to have the iconset images but much less spacing between the modules... is that possible? i tried to fiddle with `separator-block-width` in the theme, but that doesn't seem to take effect when i have an iconset setup.

update: i found the `padding` parameter in the iconset, but that's a string, and it's in the icon set... i guess i'd need to fork the awesome fonts iconset to make it do what i want?

also: if i just remove that `" "`, everything is kind of cramped together again... it's *better*, but still not ideal:

![snap-20220613T170747](https://user-images.githubusercontent.com/796623/173445708-875debda-62cb-4ffc-8c0e-d04794ba99c4.png)

sorry for being so nitpicky.. i guess it would help if there was a small diagram showing which parameters applies where. :)</div>

Now that I have an actual solution for this, it seems worthwhile to promote this to an actual bug report. ;) I hope that's alright.

PR #889, opened 2022-06-15
reduce spacing in awesome fonts iconset

anarcat:
This replaces the previous normal spacing character (which is the
usual ASCII 0x20 SPACE) by a *narrower* spacer (which is unicode
U+2009 THIN SPACE).

I found that space thanks to

https://en.wikipedia.org/wiki/Space_(punctuation)#Types_of_spaces

... and specifically:

https://en.wikipedia.org/wiki/Thin_space

... and this actually works, amazingly. Probably because it is pretty
standard as it's part of the SI specification (thousands separator),
Tex (`\thinspace` or `\,`), and HTML (`&thinsp;`)

Closes: #888

Before:

snap-20220615T115219

After: 

snap-20220615T115141

### DIFF
--- a/themes/icons/awesome-fonts.json
+++ b/themes/icons/awesome-fonts.json
@@ -1,7 +1,7 @@
 {
   "defaults": {
     "separator": "",
-    "padding": " ",
+    "padding": " ",
     "unknown": {
       "prefix": ""
     },


SamTebbs33:
Nice one :) My spaces used to look like those in the "after" screenshot and with this change they are a bit too tight together, but I expect that is something going wrong on my side. @anarcat Is your i3 configuration public so I could see what is making my spaces too narrow?

I would indeed appreciate a narrow/wide config option ;)

PR #892, opened 2022-06-22
Hide battery module if there are no batteries

SamTebbs33:
This PR hides the battery module if there are no batteries in the system. If you share a configuration between a desktop and laptop then currently the desktop will show that it's plugged in (obvious if you ask me), so with this PR the desktop doesn't show anything, which is much nicer.


tobi-wan-kenobi:
Nice, many thanks for your contribution!

PR #897, opened 2022-07-05
Updated contrib/publicip module and util/location

tfwiii:
Added another API endpoint, Added options to display country name, country code, city name and lat/long coordinates, attempt to handle failure to fetch info from API endpoints cleanly

P.S. This is my first ever pull request so apologies if I have got anything wrong!

### DIFF
--- a/bumblebee_status/modules/contrib/publicip.py
+++ b/bumblebee_status/modules/contrib/publicip.py
@@ -1,22 +1,109 @@
-"""Displays public IP address
+"""
+Displays public IP address and, optionally, Country Name, Country Code & City Name
+Maximum refresh interval should be 5 minutes to avoid free SLA breach from providers
+Note: 1 request/5 minutes is 8640 requests/month
+Provider information contained in core.location
+Left mouse click forces immediate update
+
+Parameters (Default in brackets)_
+ip (True) 		Public IP address
+country_name (False) 	Display name of country associated with the IP
+country_code (False)	Display country code of country associated with the IP
+city_name (False) 	Display name of city associated with the IP
+coordinates (False) 	Display name of city associated with the IP
+all (False) 		Display all information associate with the IP
+
+Examples
+By default only the public IP is shown
+bumblebee-status -m publicip
+
+To also include the country name...
+bumblebee-status -m publicip -p publicip.country_name=True
+
+To include all ava
+
 """
 
 import core.module
 import core.widget
 import core.decorators
-
+import core.input
+import util.format
 import util.location
 
 
 class Module(core.module.Module):
-    @core.decorators.every(minutes=60)
+    @core.decorators.every(minutes=5)
     def __init__(self, config, theme):
         super().__init__(config, theme, core.widget.Widget(self.public_ip))
 
-        self.__ip = ""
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thanks a lot, the PR looks great!

I added a few minor details, could you please address those?

Also, would it maybe be easier to just add a format string as config option? That would allow the user to specify format as well as present elements, and would probably shorten the code substantially, as well.

tfwiii:
> Thanks a lot, the PR looks great!
> 
> I added a few minor details, could you please address those?
> 
> Also, would it maybe be easier to just add a format string as config option? That would allow the user to specify format as well as present elements, and would probably shorten the code substantially, as well.
> 
> What do you think about that?

Glad it's useful!

Sorry to be dim but this is the first pull request I've ever done so I'm a little bit lost!

How do I view your comments?!

Wrt using a format string as a config option - happy to look at this. Are there any other modules that take this approach that I can use as a guide?

Thanks!

tobi-wan-kenobi:
That is an excellent question! I have actually never seen this process from a PR creator perspective. I *think* you should see my comments if you go to the "Changes" section of the PR (the diff), the comments are inlined there.

Reg. the format: The most useful example is the "disk" module, I believe, you can find a full list by going to https://bumblebee-status.readthedocs.io/en/main/modules.html and searching for ".format"

Thank you for your contribution and patience!

tfwiii:
> That is an excellent question! I have actually never seen this process from a PR creator perspective. I _think_ you should see my comments if you go to the "Changes" section of the PR (the diff), the comments are inlined there.
> 
> Reg. the format: The most useful example is the "disk" module, I believe, you can find a full list by going to https://bumblebee-status.readthedocs.io/en/main/modules.html and searching for ".format"
> 
> Thank you for your contribution and patience!

I'm sorry, I never found your comments!

But I've updated using the format string approach for parameters which I think is a big improvement.

Please let me know what you think :)

tobi-wan-kenobi:
Looks really good, thanks!

Just 2 requests (that I originally put in the vanished comments):

1. Please do not change the default interval - I think 1h is reasonable, given that public IPs, countries, etc. do not change frequently. Also, the interval can be changed configuratively.

2. Please do not change the location API (country vs country name) - I think this might be used in different modules already.

Going forward, it *might* make sense to also add a call to the location API to retrieve a dict of values, so that you don't have to maintain them manually and individually in the module. For now, it looks really good, though.

As soon as 1 and 2 are addressed, I will merge the PR.

Many thanks for your work on this!

tfwiii:
> Looks really good, thanks!
> 
> Just 2 requests (that I originally put in the vanished comments):
> 
>     1. Please do not change the default interval - I think 1h is reasonable, given that public IPs, countries, etc. do not change frequently. Also, the interval can be changed configuratively.
> 
>     2. Please do not change the location API (country vs country name) - I think this might be used in different modules already.
> 
> 
> Going forward, it _might_ make sense to also add a call to the location API to retrieve a dict of values, so that you don't have to maintain them manually and individually in the module. For now, it looks really good, though.
> 
> As soon as 1 and 2 are addressed, I will merge the PR.
> 
> Many thanks for your work on this!

Done!

(I think/hope :) )

PR #899, opened 2022-07-07
Updated to automatically update on detecting a change to default route

tfwiii:
Hello again

I've rewritten to use threading allowing updates of the widget when a change to default route is detected.

I use VPNs quite a lot with work and being able to see changes to my public IP in (near) real time is quite important. I understood that you didn't want the public geolocation API endpoints to be polled more frequently than necessary so thought this would be a good solution.

I also took the liberty of adding "contributed by `tfwiii <https://github.com/tfwiii>`_ - many thanks!" to the updated rst info - I hope that is ok!

### DIFF
--- a/bumblebee_status/modules/contrib/publicip.py
+++ b/bumblebee_status/modules/contrib/publicip.py
@@ -27,7 +27,7 @@ contributed by `tfwiii <https://github.com/tfwiii>`_ - many thanks!
 import re
 import threading
 import netifaces
-import time 
+import time
 
 import core.module
 import core.widget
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thank you very much for this contribution!

I am sick right now, so reviewing this will take a bit longer (esp.due to the changeset size). In the meanwhile: I think there are helpers for threading, but I cannot recall where exactly... Maybe in the speedtest module. Might make the code a bit easoer.

Thanks for the patience and sorry for the delay!

tobi-wan-kenobi:
Looks good to me, many thanks!

PR #905, opened 2022-07-25
Pulseaudio bug when chaning sink/source (between mono and stereo devices).

br0xpl:
Fixed a bug in pulseaudio when changing source/sink, e.g. (dis)connecting bluetooth headset.


tobi-wan-kenobi:
Nice catch, many thanks!

PR #911, opened 2022-08-09
Fix on gcalendar module

diesphink:
- Fixed save/load of token (using pickle, the other method returned error on creds.to_json() as invalid)
- Fixed for multiple calendars (was returning no event if the first calendar had no events)

### DIFF
--- a/bumblebee_status/modules/contrib/gcalendar.py
+++ b/bumblebee_status/modules/contrib/gcalendar.py
@@ -14,7 +14,7 @@ Parameters:
     * gcalendar.locale: locale to use rather than the system default.
 
 Requires these pip packages:
-   * google-api-python-client 
+   * google-api-python-client >= 1.8.0
    * google-auth-httplib2 
    * google-auth-oauthlib
 """
@@ -30,7 +30,6 @@ import core.decorators
 
 import datetime
 import os.path
-import pickle
 import locale
 
 from google.auth.transport.requests import Request
@@ -72,8 +71,7 @@ class Module(core.module.Module):
         # created automatically when the authorization flow completes for the first
         # time.
         if os.path.exists(self.__token):
-            with open(self.__token, 'rb') as token:
-                creds = pickle.load(token)
+            creds = Credentials.from_authorized_user_file(self.__token, SCOPES)
         # If there are no (valid) credentials available, let the user log in.
         if not creds or not creds.valid:
             if creds and creds.expired and creds.refresh_token:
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thank you for your contribution!

I have a question: Can we maybe store the credentials directly as binary file? Seems to me that would be easier and remove the pickle dependency.

Also, reg. regolith: The change looks good, but maybe it would be more general to use the parameter to determine a separator and use that? (I can gladly do that myself post-merge)

What do you think?

diesphink:
Hi.

> I have a question: Can we maybe store the credentials directly as binary file? Seems to me that would be easier and remove the pickle dependency.

Sure, I just don't know a reliable way to do it, pickle is already binary and is my go-to solution for those scenarios, but feel free to change anything about this.

> Also, reg. regolith: The change looks good, but maybe it would be more general to use the parameter to determine a separator and use that? (I can gladly do that myself post-merge)

Sure, just a reminder: it's the separator and also the char at the end of command sent to socket. I have no idea why the hell they changed the communication protocol on the fork.

If you make those parameters more generic, I'd also comment on the description the variation needed for regolith-rofication, because when I loaded the module it just froze the bar completely because it's waiting on the comm.

diesphink:
OH, sorry! Now I see that I've made this PR with master, instead of a branch, my mistake. That's why it duplicated the other PR.

Also, I've unintentionally added another commit with support for locale on the gcalendar

tobi-wan-kenobi:
No worries, happens :D

Do you prefer cleaning up this PR, or would you rather close this one and create a fresh one?

diesphink:
Well, the commit was another fix, to use default locale on gcalendar, and also support to set a different locale, so if your inclined to merge the other fixes, I would add this one as well.

tobi-wan-kenobi:
Good for me as well!

tobi-wan-kenobi:
OK, so,  I was basically almost merging this now, but then something occurred to me: The change from JSON to pickle is problematic to me, for 2 reasons:

1. (minor) I do not get why this is a problem. The Google API seems to offer a "credential to JSON" conversion, and that seems to fail - which is unexpected.
2. (major) How is backwards-compatibility? I.e. if somebody uses this module & has credentials stored as JSON already, how will this change affect them? Will this "auto-heal", or do they have to manually delete the credential file?

Once 2. is properly addressed, I will gladly merge the PR.

Thank you kindly for your effort!

tobi-wan-kenobi:
Great, thank you kindly!

tobi-wan-kenobi:
Changes look good to me, thank you very much!

PR #912, opened 2022-08-09
Added support for regolith fork of rofication

diesphink:
differs on the comm protocol: \n after command, receive values on csv


tobi-wan-kenobi:
I have commented in #911, which also contains this diff (accidentially?)

One more thing: Please use util.format.asbool for boolean parameters (I can adapt that post-merge)

Thank you very much for your PRs!

PR #918, opened 2022-09-08
[modules/pamixer] use -t flag to toggle mute instead of setting volume level to 0

FraSharp:
setting volume to 0 was just wrong


tobi-wan-kenobi:
great catch, many thanks!

PR #919, opened 2022-09-08
[doc] fix missing bullet on scrolling.bounce

kellya:
This is a tiny little nothing, but I noticed a missing bullet on the module doc.  This is really only prefixing a '-' on the last line of module.rst :)


tobi-wan-kenobi:
Thanks! Small improvements like this are as welcome as bigger changes - attention to detail is important for software engineers.

So: Thank you very much!

PR #923, opened 2022-09-12
Added task detail display for taskwarrior

kellya:
Updated taskwarrior.py to accept a show.active param, which will change the behavior to show pending tasks ordinarily, but will display the current active task ID and description if the option is set, and there is an active task.


tobi-wan-kenobi:
very nice addition, many thanks!

PR #926, opened 2022-09-15
[modules/taskwarrior] add state handling

kellya:
Sets one of two states:
  "stopped" - Default, no running task
  "active"  - When an active task is running.


tobi-wan-kenobi:

PR #927, opened 2022-09-16
Correct battery modules handling of critical/warning states

kellya:
The logic for the critical/warning handling on the battery modules was applied BEFORE the discharging-<pct> logic.  This made it  possible for those discharging states to get applied over a critical/warning and allow a theme to override the critical/warning colors with a state of "discharging-10", for example.

This change moves that logic after the discharging state, so that it will always "win" if critical or warning states are set.

This also adds the "discharging" battery state to the critical/warning check so the state will apply if the battery is not on AC power, but would return normal otherwise.  Meaning, if the battery is "critical" from a percentage check, but is plugged into power, the critical state is removed.


tobi-wan-kenobi:
Ah, that makes a lot of sense, many thanks!

Note to self: Tests for that particular case would be nice!

PR #929, opened 2022-09-20
contrib/network_traffic dependency in docs

alonsomoya:
network_traffic [is dependent on netifaces](https://github.com/tobi-wan-kenobi/bumblebee-status/blob/main/bumblebee_status/modules/contrib/network_traffic.py#L11) which is not a core python module. Reflecting it on the docs


tobi-wan-kenobi:
Thank you kindly for your contribution!

PR #930, opened 2022-09-21
Don't install manpages to /usr/usr

benthetechguy:
`data_files` shouldn't have `usr/` in it; this causes the manpages to be installed to `/usr/usr/share/man/man1` instead of `/usr/share/man/man1`.


tobi-wan-kenobi:
never noticed that. thank you very much for the PR!

PR #933, opened 2022-10-06
publicip module - Fixed bug and minor improvements in output

tfwiii:
Handled fail where core.location does not provide values for latitude and longitude. Added handling for coordinates N, S, E, W.

### DIFF
--- a/bumblebee_status/modules/contrib/publicip.py
+++ b/bumblebee_status/modules/contrib/publicip.py
@@ -60,43 +60,34 @@ class Module(core.module.Module):
         self.__monitor.start()
 
     def monitor(self):
-        default_route = None
-        interfaces = None
+        __previous_ips = set()
+        __current_ips = set()
         # Initially set to True to force an info update on first pass
-        information_changed = True
+        __information_changed = True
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
thank you very much for the PR!

I addes 2 small remarks in the code directly, once we address those, the PR is ready to be merged, from my PoV.

tfwiii:
Can't see you comments but presume they related to the debugging prints I had left in the code. Apologies for that, they are now removed!

tobi-wan-kenobi:
thanks! exactly :)

the other thing i don't understand is line 129. to my understanding, __lon and __lat are never initialized, am i missing something? so basically, it wozld always display unknown, because the fields are not extracted from __info.

maybe it is just the github formatting, i am on mobile right now.

tfwiii:
I think you're right! I'm sorry, I rushed this before going out and I obviously didn't test it enough. Let me take another look when I get home and I'll update...

tobi-wan-kenobi:
Looking good now, thank you very much for your contribution!

PR #934, opened 2022-10-12
publicip - Bug Fix

tfwiii:
IP address changes were being missed if an interface was present but did not have an IPv4 address associated with it. Added exception handling to mitigate this.


tobi-wan-kenobi:
thank you very much, looks good!

PR #942, opened 2022-11-29
`watsonctl.py` an interactive extension to `watson.py`

Dale-Muccignat:
I'm hoping this will serve as an alternative to watson.py. I started using watson.py and felt like it'd be great to have some control over what project is being tracked rather than 'just' restarting the last tracked project.

So I went about changing the code from watson.py to include a feature that when the mouse wheel is used, the module cycles between currently used projects within watson. I have limited python experience as I mainly use Julialang for research so I'm reaching the end of the effort I can put into this for now. I'd love for some pointers and I can come back to improve this.

I'm not really looking to merge this right away, I think it's usable as is but I would love for it to be improved. 

Main features:
- Mouse scroll to select from existing projects

Current drawbacks:
- No way of starting a new tracked project. I can't wrap my head around a way to easily start a new project with a custom name from the module
- No fancy icons, I need to read more of the docs for bumblebee to add in some fancy icons to indicate if the project is being tracked or not currently

Wishlist:
- right-click to display the aggrigate total times for each project
- change colour of the module when time is being tracked

### DIFF
--- a/bumblebee_status/modules/contrib/watsonctl.py
+++ b/bumblebee_status/modules/contrib/watsonctl.py
@@ -1,0 +1,92 @@
+# pylint: disable=C0111,R0903
+
+"""Displays the status of watson (time-tracking tool)
+Requires the following executable:
+    * watson
+origional module contributed by `bendardenne <https://github.com/bendardenne>`_ - many thanks!
+extended by `dale-muccignat <https://github.com/dale-muccignat>`_
+"""
+
+import logging
+import re
+import functools
+
+import core.module
+import core.widget
+import core.input
+import core.decorators
+
+import util.cli
+
+
+class Module(core.module.Module):
+    # @core.decorators.every(minutes=1)
+    def __init__(self, config, theme):
+        super().__init__(config, theme, core.widget.Widget(self.text))
+
+        self.__tracking = False
+        self.__status = ""
+        self.__project = "Select Project"
+
+        self.__project_key = {}
+        self.__project_list = []
+        self.get_list()
+
+        core.input.register(self, button=core.input.LEFT_MOUSE, cmd=self.toggle)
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thank you very much!

Regarding the new project start: Tricky, probably needs a bumblebee-status core extension that shows a prompt (Tk? Gtk? Qt? Just a term?) I think there are x tools to show a prompt and read the result, maybe that is a good starting point.

The "use last project": Good idea, would probably mean introducing the concept of "state" to bumblebee, which IMHO sounds very useful :)

Unfortunately, I have very limited time to work on bumblebee right now (real life interfering), and I do not use watson myself, but I will try to think about those two things and come up with ideas or prototypes.

Thank you!

Dale-Muccignat:
Text prompt: Yeah I'm not too sure myself, I'll have a think over time for some ideas around inputting text. I'm not sure if it'd have any applications for other modules but could be a cool feature if it's easy enough to do.

State: Oh I hadn't thought about it like that, but it sounds like another toy that 'could' be useful. I was more imagining that I could probably figure out which was the last used project from watson itself. Shouldn't be tooo hard.

I'm the same with time, I'm trying to get a bunch of work done (hence seeking out some time tracking tools.... and then spending a whole day trying to tailor it to my wants... very productive!). I'll most likely just see if I stick with using it and find it useful, then when I have some spare time I'll try and come up with ideas.

Thanks for chiming in!

Dale-Muccignat:
I've added a simple (but ugly) way of entering text. On right-click, it opens up a gui from the python package "easygui". 

I don't particularly like that it's a whole new window but it's a solution for now. Having a more general solution would probably be preferred.

tobi-wan-kenobi:
Very cool!

Can you please drop me a line when you think it is ready for merging?

PR #943, opened 2022-11-30
Documentation: Fix the default format string for nvidiagpu

pvutov:
Was surprised to get errors copy-pasting from the manual, until I looked at the list of legal parameters :)


tobi-wan-kenobi:
Oh, good catch! Many thanks!

PR #947, opened 2023-01-04
Added a max_chars parameter to be able to control the widget width.

arivarton:
Also moved the try block a bit further up to catch network errors.

### DIFF
--- a/bumblebee_status/modules/contrib/gcalendar.py
+++ b/bumblebee_status/modules/contrib/gcalendar.py
@@ -3,7 +3,9 @@
 Events that are set as 'all-day' will not be shown.
 
 Requires credentials.json from a google api application where the google calendar api is installed.
-On first time run the browser will open and google will ask for permission for this app to access the google calendar and then save a .gcalendar_token.json file to the credentials_path directory which stores this permission.
+On first time run the browser will open and google will ask for permission for this app to access
+the google calendar and then save a .gcalendar_token.json file to the credentials_path directory
+which stores this permission.
 
 A refresh is done every 15 minutes.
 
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thanks a lot to the PR!

The "catching" changes look totally fine for me, regarding the max width: Would it maybe be easier to just give the module the "scrollable" decorator? That way, users can provide a width, and still see all content because the widget scrolls.

Alternatively, I think I would prefer a generic solution ("max-width", analogous to the existing "min-width")

I generally try to stear clear from options that change the size of a widget frequently, if possible, because I feel it makes the status bar very noisy.

Thank you very much again for giving your improvements back, I am looking forward to your thoughts on my suggestions above.

arivarton:
I was experimenting with the scrollable decorator but didn't really get it to work as expected. 
Another issue there is that the date and time will scroll away when its moving, but I guess I can add the date and time in a separate widget and only scroll the text.

I will take another look at this.

arivarton:
> Thanks a lot to the PR!
> 
> The "catching" changes look totally fine for me, regarding the max width: Would it maybe be easier to just give the module the "scrollable" decorator? That way, users can provide a width, and still see all content because the widget scrolls.
> 
> Alternatively, I think I would prefer a generic solution ("max-width", analogous to the existing "min-width")
> 
> I generally try to stear clear from options that change the size of a widget frequently, if possible, because I feel it makes the status bar very noisy.
> 
> Thank you very much again for giving your improvements back, I am looking forward to your thoughts on my suggestions above.

I've made some changes to this pull request which should satisfy your specifications.

tobi-wan-kenobi:
Looks good to me - and good catch that scrolling overrides the "every" specifier - I think that is a bug, actuall,.

Will look into that. Thank you!

PR #949, opened 2023-02-04
add pipewire module

jebaum:
---
name: add module for pipewire
about: uses wpctl to get and update volume in pipewire
---

### Description

**What kind of change does this PR introduce?**
add a module for pipewire. basically identical to existing amixer and pamixer modules, but using wpctl

**Summary**
for folks using pipewire, they probably have a pipewire pulse bridge of some kind installed that some applications depend on, and pamixer/pactl/other pulse things may work intermittently depending on which application is currently playing audio. wpctl should work all the time though

**Does this PR introduce a breaking change?**
no

**Other information**
I tested this and it Works On My Machine (tm)

### DIFF
--- a/bumblebee_status/modules/contrib/pipewire.py
+++ b/bumblebee_status/modules/contrib/pipewire.py
@@ -1,0 +1,90 @@
+"""get volume level or control it
+
+Requires the following executable:
+    * wpctl
+
+Parameters:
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
great addition, thank you very much!

PR #950, opened 2023-02-10
fix bug in pipewire module

jebaum:
- fix bug in pipewire module

---
name: Improvement
about: You have some improvement to make bumblebee-status bar better?
---

### Improvement

there was a bug in my initial implementation - if the string "N/A" is returned because for some reason the volume can't be fetched for a tick during an update (seems to happen sometimes when switching audio streams), the code tries to convert it to a float, which throws a ValueError and crashes bumblebee. make sure to catch that case


tobi-wan-kenobi:
Many thanks!

PR #964, opened 2023-04-13
Change OpenWeatherMap request url from HTTP to HTTPS

dmturner:
I was noticing that the weather module had stopped working and was hanging waiting for a response from http://api.openweathermap.org.

Changing this request to "https://" fixed the issue and the weather module works again.


tobi-wan-kenobi:
whoopsie :)

good change, thank you!

PR #968, opened 2023-05-06
Updated gitlab module to have state of warning when there is at least 1 notification, just like the github module

Duarte-Figueiredo:
In order to mimic the same behavior as the github module, I've update the gitlab module to have a state which will turn to warning if there is one or more notification detected


tobi-wan-kenobi:
very nice, thank you very much!

PR #972, opened 2023-05-28
use API token instead of password hash in pihole module

bbernhard:
* with newer versions of pi-hole, it is not possible anymore to use the password hash for the API authentication. Instead, one needs to use the dedicated API token for that. In order to stay backwards compatible, and not break existing bumblebee_status setups, the 'pwhash' parameter is still supported (in case someone runs an outdated pi-hole version).

* the new pi-hole API endpoints do not allow to access the summary endpoint without an API token. So, therefore '&auth=<api token>' was added.


tobi-wan-kenobi:
thanks a lot!

PR #973, opened 2023-06-19
[modules/cpu3] Add new CPU module

SuperQ:
Based on cpu2 module, but use `sensors -j` and some json path walking to better parse CPU temp and fan speeds. The output of `sensors -u` can contain many duplicates of the same sensor name, but from different sensor device paths.


tobi-wan-kenobi:
Thank you kindly for the contribution!

PR #976, opened 2023-07-10
[modules/usage] A module for "ActivityWatch"

lasnikr:
Essentially, it's an enhanced version of the `uptime` module with additional dependencies.  
It directly accesses the local database of [ActivityWatch](https://activitywatch.net/). I'm not sure if this is recommended, as there isn't much comprehensive documentation available for it.

However, since it shouldn't cause any harm, apart from the possibility of the module becoming outdated at some point (which can be resolved by setting a new default for "usage.database"), it utilizes the local database.


tobi-wan-kenobi:
nice, thank you very much!

PR #977, opened 2023-07-16
allow setting sink ID in pipewire module

jebaum:
### Improvement

currently the pipewire module only works with whatever the default audio sink is
if you use multiple sinks or easyeffects, this may not let you achieve the behavior you want
allow passing a sink ID, and fall back to the current behavior if no argument is passed


jebaum:
not quite sure why the tests are failing, at first glance it doesn't look related to the PR content

tobi-wan-kenobi:
Changes look good, thank you kindly for the PR!

Yeah, the tests fail since some time, unfortunately, didn't have time to investigate yet....

PR #979, opened 2023-07-19
Add: missing mention of pipewire module on website

sazk07:
I was reading the docs on the website(https://bumblebee-status.readthedocs.io/en/latest/modules.html#contrib) and there was no mention of the pipewire module which exists among the contrib modules. This edit will inform users about the pipewire module


tobi-wan-kenobi:
many thanks for the PR!

PR #980, opened 2023-07-19
Fix: pipewire.py module

sazk07:
mousewheel up and down events were initially not working on my bumblebee status bar. After verifying on the command line, it seems that wpctl set-volume command requires sink ID first and percentage change as the second arg.

steps to verify, assuming wireplumber is installed and sink ID is 32

```sh
wpctl set-volume 32 50%
```
will set the volume to 50%

```sh
wpctl set-volume 50% 32
```
will result in `Object '52' not found`

After switching `self.__change` and `self.__id`, mousewheel up and down events worked as expected increasing and decreasing volume. 

I also wanted to take the opportunity to ask you if we can write python modules using f-strings instead of the str.format() for better readability? I believe this will break compatibility with Python versions <3.6 as f-strings were introduced in Python v3.6.

### DIFF
--- a/bumblebee_status/modules/contrib/pipewire.py
+++ b/bumblebee_status/modules/contrib/pipewire.py
@@ -57,12 +57,12 @@ class Module(core.module.Module):
 
     def increase_volume(self, event):
         util.cli.execute(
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Thank you very much for the contribution and good find!

PR #986, opened 2023-08-30
Allow adjusting the fontsize of the pulsectl default device popup

TheEdgeOfRage:
I've recently started using the pulsectl tk popup to pick the default device and one annoying thing is that the default font size is way too small for me, so I added basic support for changing it. I'm not sure if this is a good way to implement it, let me know if not :)

Edit: I just remembered that I should probably also update the docs :sweat_smile: I'll try to get around to that later today after work

### DIFF
--- a/bumblebee_status/modules/contrib/bluetooth.py
+++ b/bumblebee_status/modules/contrib/bluetooth.py
@@ -75,19 +75,15 @@ class Module(core.module.Module):
 
     def popup(self, widget):
         """Show a popup menu."""
-        menu = util.popup.PopupMenu()
+        menu = util.popup.menu(self.__config)
         if self._status == "On":
-            menu.add_menuitem("Disable Bluetooth")
+            menu.add_menuitem("Disable Bluetooth", callback=self._toggle)
         elif self._status == "Off":
-            menu.add_menuitem("Enable Bluetooth")
+            menu.add_menuitem("Enable Bluetooth", callback=self._toggle)
         else:
             return
 
-        # show menu and get return code
-        ret = menu.show(widget)
-        if ret == 0:
-            # first (and only) item selected.
-            self._toggle()
+        menu.show(widget)
 
     def _toggle(self, widget=None):
         """Toggle bluetooth state."""
--- a/bumblebee_status/modules/contrib/system.py
+++ b/bumblebee_status/modules/contrib/system.py
@@ -8,11 +8,11 @@ adds the possibility to
 	* reboot
 
 the system.
-	
+
 Per default a confirmation dialog is shown before the actual action is performed.
-	
+
 Parameters:
-	* system.confirm: show confirmation dialog before performing any action (default: true) 
+	* system.confirm: show confirmation dialog before performing any action (default: true)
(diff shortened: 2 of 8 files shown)


bumblebee_status/modules/core/pulsectl.py
```diff
@@ -65,6 +65,9 @@ def __init__(self, config, theme, type):
         self.__show_device_name = util.format.asbool(
             self.parameter("showdevicename", False)
         )
+        self.__popup_font_size = util.format.asint(
+            self.parameter("popup_font_size", None), 0
+        )
```
TheEdgeOfRage, line 70: For now, I've only added the font size parameter to this module, but it should probably be added to all the ones that use the popup menu

tobi-wan-kenobi:
Nice addition, thank you very much! Great to read from you again!

I think eventually, the best option probably would be having this as global (module independent) parameter. What do you hink?

TheEdgeOfRage:
Oh, I didn't know you can have global ones. That would definitely be preferred then :D

How do I define it globally?

tobi-wan-kenobi:
Those are the parameters defined outside the modules (in config.py, i think, can't remember for sure and don't have a pc handy)

I *think* popup should then have access to that via a config object.

tobi-wan-kenobi:
Looking at it in a bit more detail:

* If you look at bumblebee_status/core/config.py, you can see how "interval" is set on a global level, and use it via `config.<accessor>()`. I think that maybe this would be the best option here.
* More generally, it might make sense to always recurse to a global parameter, if no more specific one is present (i.e. use `pulsein.fontsize`, but if that is not found, use `fontsize` - what do you think?
* Generally, I think the fontsize logic should be in the popup module, not in the individual widget modules, what do you think?

If you prefer, I can put together a draft PR for those changes, as they are targeting the core quite a bit.

TheEdgeOfRage:
Sorry, I haven't had much time these days to finish this. I've removed the code from the pulsectl module and I agree with your third point that it should ideally all be in config and the popup module. The issue I'm having is that config is a Class definition, but the instance is only available in the main file. Would it make sense to have config be a singleton that can be imported from other modules?

bumblebee_status/modules/contrib/bluetooth.py
```diff
@@ -75,19 +75,15 @@ def update(self):
 
     def popup(self, widget):
         """Show a popup menu."""
-        menu = util.popup.PopupMenu()
```
TheEdgeOfRage, line 78: This seems to have been using an old implementation, so I've updated it

bumblebee_status/modules/contrib/system.py
```diff
@@ -77,7 +77,7 @@ def popup(self, widget):
             util.cli.execute(popupcmd)
             return
 
-        menu = util.popup.menu()
+        menu = util.popup.menu(self.__config)
```
TheEdgeOfRage, line 80: This is the best I came up with for passing on the config object to the popup menu. WDYT?

tobi-wan-kenobi:
Looks good, thank you very much!

PR #987, opened 2023-09-06
Update introduction.rst

PrabhuUdurg:
I faced with pretty stupid error because of this, so I reckon it would be helpful.

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -43,6 +43,7 @@ like this:
            -p <list of module parameters> \
            -t <theme>
    }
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
What was the problem exactly? Newlines without the \ at the end?

I just want to make sure to improve the docs as much as possible.

PrabhuUdurg:
My problem was that after the backslash I started new line as it was in the example in documentation, but it's caused an error, which I fixed putting every -m values in one line.

tobi-wan-kenobi:
OK, that is weird, because escaped newlines *should* work. Could you maybe paste the broken config?

PR #991, opened 2023-09-13
poll window title also on workspace change (not only window events)

LokiLuciferase:
Currently, the `title` module only considers window events. That means that the previous window title remains displayed even if the workspace is switched to an empty workspace. 

This changes that behaviour so that also workspace events are listened to. Now switching to an empty workspace removes the window title and replaces it with the name of the workspace.


tobi-wan-kenobi:
Oh, that is a really good change, thank you kindly!

PR #994, opened 2023-09-20
feat: add `disabled` parameter

zetxx:
it pauses dunst notification on startup


tobi-wan-kenobi:
Ah, nice addition, thank you!

PR #996, opened 2023-10-01
Module for displaying date with gnome-calendar opening on click

surbhitt:
Hi, 
I love the project been using it for a while. I love how easy it is to use and even implement custom modules. Thankfully to that I have been able to seamlessly modify it for my use case. I felt a personal requirement for this module as I need to open up calendar for setting up reminders a lot and opening it up from the bar seemed the easiest and most natural, probably **this should exist in the core date module** but didn't wanna touch it.
Let me know if I should make any changes to this or anything at all, I am new to all this... I appreciate the feedback/critic. :)

### DIFF
--- a/bumblebee_status/modules/contrib/calendar.py
+++ b/bumblebee_status/modules/contrib/calendar.py
@@ -1,0 +1,24 @@
+"""Displays date, opens a gnome-calendar on click
+
+Parameters:
+    * calendar.format: strftime()-compatible formatting string
+    * calendar.locale: locale to use rather than the system default
+"""
+
+import core.decorators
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Thanks a lot for the kind words, for using bumblebee and for the PR!

This actually does exist in the core tool itself, but is possibly not so well documented: *every* module has a "left-click" property to add actions on click, like this:

`bumblebee-status -m date -p date.left-click=gnome-calendar`

This will execute gnome-calender when left-clicking on the module.

Hope this works for you and helps!

surbhitt:
thank you so much that is what I was looking for.

PR #999, opened 2023-10-02
Pomodoro Audio Notifications 

PrabhuUdurg:
Hi! Most Pomodoro timers have a sound notification, which is really helpful. I added this feature also to the Pomodoro module. I used simpleaudio library to play sounds. Also, I added two variables at the beginning of the file, for the work and rest duration, to make the user change it easier, and for better maintainability.
Additionally, I added a sentence about how to change these parameters in the header. 
Thank you


tobi-wan-kenobi:
Thanks a lot for the PR!

I have a couple of small suggestions/requests:

1. Please keep the 2 parameter hardcoded, those are just the default values and can be changed using "-p pomodoro.<name>". Users should *not* have to modify the source code (because updates will override their changes)

2. Instead of committing the wav, please let users specify (via parameter) the audio file they want to play. I don't know where that wav came from and want to avoid copyright/usage dispute problems. Any content not directly generated by bumblebee has to come from the user.

Hope that helps! Thank you!

PrabhuUdurg:
Hallo! I can remove the path and file, but then the library throws an error. How I should handle this, thank you

tobi-wan-kenobi:
If the wav does not exist, I would suggest to just log an error (using log.error) and continue.

PrabhuUdurg:
Hallo! I made changes, now user can upload their own audio files, The error problem was handled by try except.

Issue #821, opened 2021-09-28
Package bumblebee-status for nixos


Deep-Six:
### Feature Request
It'd be great to have bumblebee-status within the NixOS ecosystem, I'm relatively new to NixOS so am unfortunately familiar enough to support this.

tobi-wan-kenobi:
Yes, that would be great! Unfortunately, I have never used NixOS, so I have no idea how to package for it.

Deep-Six:
https://www.youtube.com/watch?v=jXd-hkP4xnU This is something I've tried to grok, but it's a bit much for me at this time. Maybe it'd help you?

tobi-wan-kenobi:
Nice video :+1: 

I've also started reading https://nixos.wiki/wiki/Packaging/Tutorial and https://nixos.org/manual/nixpkgs/stable/#chap-quick-start, but frankly, it seems this goes way beyond the time I am - unfortunately - able to dedicate to this right now.

I have tagged the issue with "help wanted", so maybe somebody with more NixOS knowledge can jump in.

Thanks a lot for raising the issue!

Deep-Six:
@tobi-wan-kenobi  I have managed to learn enough to stumble through the packaging process, but in the process I had to patch the themes path, is there a reason why you're not using the $XDG_DATA_DIRS convention for the themes?

tobi-wan-kenobi:
The only reason I can bring forth is missing knowledge on my part. Does make a lot of sense to use that convention.

I will add that, thanks for raising it! (might take a while though, due to Covid messing up my schedule quite a bit roght now)

Deep-Six:
Ironically it was COVID (recovery) that gave me the free time to finally dig into this. Stay safe!

Tonus1:
> The only reason I can bring forth is missing knowledge on my part. Does make a lot of sense to use that convention.
> 
> I will add that, thanks for raising it! (might take a while though, due to Covid messing up my schedule quite a bit roght now)

Recover soon and well !

tobi-wan-kenobi:
@Deep-Six Does this change work for you? I tested it locally and it seems to do what is intended, but I am not 100% sure the subdirectories (bumblebee-status/themes/ are what you need.

Deep-Six:
@tobi-wan-kenobi  I can work with it, I need to also patch up the modules path, is there an equivalent to PATHS for modules?

tobi-wan-kenobi:
[...]

PR #1003, opened 2023-11-29
Added instructions for nixpkgs install

adgai19:
Followup to https://github.com/tobi-wan-kenobi/bumblebee-status/issues/821.


tobi-wan-kenobi:
Nice, thank you very much!

PR #1004, opened 2023-11-30
[modules/disk] Use bavail instead of bfree

czert:
The ext4 filesystem reserves 5 % of disk size. `os.statvfs(path).f_bfree` reports this reserve as free disk space, while `f_bavail` considers it "full".

I believe `f_bavail` gives more useful information, as the reserved space cannot be used. Also, it's in line with what `df` reports.


tobi-wan-kenobi:
Sounds sensible, many thanks!

PR #1005, opened 2023-12-06
redshift: added -P argument to prevent cumulative effects

alexveden:
In adjust=true mode, redshift starts to stack applied effect and screen starts dim more and more every interval. Adding -P argument solves this issue.


tobi-wan-kenobi:
Nice, thank you ver much!

PR #1006, opened 2023-12-10
Layout: highlights when layout is not in default state

alexveden:
This pull request introduces a new option for highlighting non-default layout:
```
layout.highlight_if_not = us
```

Does the following:
Default layout
image

Non-default;
image

Previous plugin behavior preserved, if option not set, no highlighting for both layouts.

### DIFF
--- a/bumblebee_status/modules/contrib/layout.py
+++ b/bumblebee_status/modules/contrib/layout.py
@@ -5,7 +5,11 @@
 Requires the following executable:
     * setxkbmap
 
+Parameters:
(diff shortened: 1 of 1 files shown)


alexveden:
BTW, when I was working on layout module, it seems there are 2 `layout.py`  in core and contrib dirs. Bumblebee-status picks the contrib verison, I don't know if this is an issue or intentional. Maybe more clarity is required in the docs.

tobi-wan-kenobi:
Oh, I did not notice the issue regarding duplicate module names, I will certainly investigate that. Thanks for mentioning this and thanks for the cool PR!

tobi-wan-kenobi:
This might be functionality that could be nice in the framework itself - add a highlight group if a widget does (not) have a certain value.

PR #1013, opened 2024-03-07
fix(modules,contrib,thunderbird): grep binary file error

DrBu7cher:
I saw that grep sometimes thinks the `.msf` file is a binary, so I added `-a` as a flag which tells grep to look inside binary files as well.
From `man grep`:
```
   File and Directory Selection
       -a, --text
              Process a binary file as if it were text; this is equivalent to the --binary-files=text option.
```


tobi-wan-kenobi:
Thank you!

Issue #1014, opened 2024-03-19
Sun module throws error into the bar : "error: can't compare offset-naive and offset-aware datetimes"


UnchartedSystems:
![image](https://github.com/tobi-wan-kenobi/bumblebee-status/assets/20145459/cb9bbe52-bedc-4aa2-9c63-dde4e7d7a288)

Dependencies installed:
requests : 2.28.2
python-dateutil: 2.9.0-post0
suntime: 1.3.2

I'm running Fedora 39 i3 spin. I've tried both using local and global time on my system.

dmturner:
@UnchartedSystems did #1017 manage to fix this issue?

tobi-wan-kenobi:
closing due to inactivity, thanks a lot to @dmturner for jumping in!

PR #1017, opened 2024-04-13
Fix for deprecated functions throwing error

dmturner:
The "sun.py" module was throwing an error as described in #1014. This fix replaces the deprecated function `get_local_sunrise_time()` with `get_sunrise_time()`.


tobi-wan-kenobi:
thanks a lot!

PR #1035, opened 2025-01-05
Add oled_offset module

TheEdgeOfRage:
I recently got an OLED monitor and as a precaution, decided to add this module that occasionally (every minute) moves the entire bumblebee-status status bar to the left, so that it's not always the same pixels that are lit up, to mitigate burn-in. It resets back to 0 after 3 steps.


tobi-wan-kenobi:
Nice idea, thank you very much! I wasn't aware that burn-in is (still) an issue for OLED, that's interesting to know.

TheEdgeOfRage:
It's mostly fine these days with new monitors, but I spend hours and hours with my bumblebee modules pretty much being static, so I wanna be extra careful :sweat_smile:

PR #1037, opened 2025-02-07
Fix the parameter name for the `pipewire` module

Fittiboy:
The current `wpctl.percent_change` is not valid, while `pipewire.percent_change` does work!


tobi-wan-kenobi:
Oohh, good catch, thank you!

PR #1038, opened 2025-02-22
updated pi-hole integration to v6 API

bbernhard:
With the release of pi-hole v6, the developers completely changed the API and how one interfaces with it. This changeset makes the pihole integration in bumblebee-status work with the new API.

Be aware: This is a backwards incompatible change. i.e the integration doesn't work with the v5 API anymore! Since pi-hole v6 is the way forward and people will eventually move to the new release, I think it doesn't make sense to keep the old pi-hole support alive. Getting rid of the old code also reduces the LOC in the integration and makes it easier to maintain :)


tobi-wan-kenobi:
sounds good, many thanks! very cool addition to bumblebee!

PR #1040, opened 2025-03-04
Epoch Timestamp Module

theymightbetim:
Adds a module that displays the current time as an epoch timestamp and copies it to the clipboard when you left click on it.

### DIFF
--- a/bumblebee_status/modules/contrib/epoch.py
+++ b/bumblebee_status/modules/contrib/epoch.py
@@ -11,16 +11,22 @@ import core.input
 
 import time
 
+
 class Module(core.module.Module):
-    
+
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi:
Nice, thanks!

I thought that would be possible with the current time module, but turns out, I am wrong. This is really good functionality for developers. Thank you!

theymightbetim:
> Nice, thanks!
> 
> I thought that would be possible with the current time module, but turns out, I am wrong. This is really good functionality for developers. Thank you!

and QA Engineers ;)

tobi-wan-kenobi:
:-D Absolutely!

PR #1041, opened 2025-03-07
feat: OTP module

theymightbetim:
- Adds a new module which can generate and display a One Time Pin for Two Factor Authentication using a secret key stored in keyring.
 - Left click to copy to clipboard with xclip


tobi-wan-kenobi:
Thanks! looks almost like "copy contents to clipboard" might become useful as a general-purpose utility.

PR #1043, opened 2025-03-07
fix: blank line at end of requirements file to fix module imports

theymightbetim:
I broke the test pipeline again with the OTP module because the requirements file was missing a blank line at the end causing it to concatenate the last module in otp.txt with the first module in pihole.txt and then become unable to find a package called pyotprequests instead of installing pyopt and requests.

Sorry about this, I was unable to run the tests before because the pipeline was broken.

[Tests Passing Again](https://github.com/theymightbetim/bumblebee-status/actions/runs/13722033184)

### DIFF
--- a/requirements/modules/otp.txt
+++ b/requirements/modules/otp.txt
@@ -1,3 +1,2 @@
 keyring
 pyotp
-requests


requirements/modules/otp.txt
```diff
@@ -1,3 +1,2 @@
 keyring
 pyotp
-requests
```
theymightbetim, line 3: This package isn't actually required by this module. We need a blank line at the end of the file to prevent concatenation with the next file's requests module.

theymightbetim:
> thanks for your efforts!

Just happy to see all tests passing again :)

tobi-wan-kenobi:
> Just happy to see all tests passing again :)

me too, me too :-)

PR #1045, opened 2025-03-23
fix: dnf module doesn't work for latest version of fedora/dnf5

theymightbetim:
The current implementation of the dnf module doesn't work for Fedora 41. 
`dnf updateinfo` outputs a missing argument error causing it to always display 0/0/0/0

-Updated command to dnf updateinfo summary

the output of dnf updateinfo summary differs slightly from the former format. The new format is as follows:
```.venv➜  bumblebee-status git:(fix-dnf) ✗ dnf updateinfo summary         
Updating and loading repositories:
Repositories loaded.
Available advisory information summary:
Security    : 28
  Critical  : 1
  Important : 7
  Moderate  : 15
  Low       : 3
  Other     : 2
Bugfix      : 90
Enhancement : 63
Other       : 46
````

- Updated the parsing function to handle the new format.

### DIFF
--- a/bumblebee_status/modules/contrib/dnf.py
+++ b/bumblebee_status/modules/contrib/dnf.py
@@ -30,14 +30,14 @@ class Module(core.module.Module):
 
     def update(self):
         widget = self.widget()
-        res = util.cli.execute("dnf updateinfo", ignore_errors=True)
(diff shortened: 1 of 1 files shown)


tobi-wan-kenobi (approved):

tobi-wan-kenobi:
Thanks a lot for the fix!

PR #1046, opened 2025-03-26
Popup calendar module

theymightbetim:
Extended version of datetime module which can display calendar on click and open google calendar on the selected date if you hit enter.

popupcal


tobi-wan-kenobi:
Very nice!!!

One suggestion: I think "calendar" would be a nicer name for it, as people would more readily know what it does. What do you think about renaming it?

If we can make it backwards compatible, I'd even consider this as a datetimee replacement (but "calendar" might be the better way to go about it)

Thank you!

theymightbetim:
@tobi-wan-kenobi did the rename you suggested and cleaned this up a bunch and added tests. should be ready now.

tobi-wan-kenobi:
Looks good, many thanks!

PR #1047, opened 2025-03-27
feat: Scratchpad module

theymightbetim:
This module displays a count of windows currently on the scratchpad and allows you to launch a window picker using rofi by clicking on the module.


theymightbetim:
Converting back to draft because I just can't get the tests to work. Either theres not enough coverage or when there is, on the pipeline it won't actually get i3 data back so it fails.

PR #1050, opened 2025-05-19
Updated git module to support drawing widgets right to left

morgabm:
I am a first time user of bumblebee-status and was configuring it when I noticed there was support for rtl layout of modules, but not for individual widgets within many of the modules. 

Primarily, I would like to be able to use the git module in this way, so here is my PR.

# What this does
* Add support for rtl layout of widgets within git module
* Maintain support for ltr layout of widgets within git module
* Allows users to configure layout direction within the git module with parameter `git.draw_order`

# What this doesn't do
* Add support for rtl layouts in other modules
* Change behavior of any usage currently supported
* Change the order of the widgets with git module (still git.new -> git.modified -> git.deleted in either direction)

# Example usage
`bumblebee-status -t dracula-powerline -r -m date time battery pipewire xrandr nic git -p nic.format="{ssid}" nic.states="up" date.format="%a, %b %-d %Y" time.format="%-I:%M %p" git.draw_order="rtl"`
2025-05-19-174247_872x44_scrot

### DIFF
--- a/bumblebee_status/modules/core/git.py
+++ b/bumblebee_status/modules/core/git.py
@@ -6,6 +6,9 @@ currently focused window.
 Requires:
     * xcwd
     * Python module 'pygit2'
+
+Parameters:
+    * git.draw_order: String to specify draw order of the widgets; Options are "ltr" for left to right, and "rtl" for right to left (defaults to "ltr")
 """
 
 import os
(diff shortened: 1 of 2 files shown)


tobi-wan-kenobi:
Sorry, took me a while to review this. Looks good, thank you!

PR #1051, opened 2025-05-19
Updated dracula theme with styling for more modules

morgabm:
I extended the shipped with dracula theme a bit and figured I would offer it here.


tobi-wan-kenobi:
Thank you very much!

PR #1066, opened 2026-02-10
Add multiple paths support to the disk module + showpath" config option + mouse wheel scroll

cirsm:
This change allows users to:
-Specify multiple mount points or paths via the path option
-Toggle absolute (full) path display using the new boolean showpath option in the i3wm config
-Scroll between the selected paths using the mouse wheel


tobi-wan-kenobi (approved):

tobi-wan-kenobi:
great addition, thank you!

PR #1068, opened 2026-02-12
fix(disk): correct left-click to open selected directory and remove redundant showpath option

cirsm:
Fixed left-click action to now open the currently selected/visible mount point (instead of always the first one)
Removed the now-unnecessary "showpath" config option (the format option already does what this option did)


tobi-wan-kenobi (approved):

tobi-wan-kenobi:
thanks!

PR #1069, opened 2026-02-16
feat(nic): support custom command on left-click via nic.open

cirsm:
Default: '/bin/xterm nmtui'


tobi-wan-kenobi:
afaicr, *all* modules have abgeneric "left-click" attribute that should already allow you to do that.

is that broken, or just badly documented?

https://bumblebee-status.readthedocs.io/en/main/features.html#events

cirsm:
Thanks! It does work as intended, my mistake!

